PR #6, opened 2025-02-17
Prune half-finished and untested code

krystophny:


src/neort.f90
```diff
@@ -0,0 +1,420 @@
...
+    end subroutine init_plasma
+
+    subroutine init_run(use_thermodynamic_profiles)
+
+        logical, intent(in) :: use_thermodynamic_profiles
+
+        call init
+
+        Om_tE = vth*M_t/R0                   ! toroidal ExB drift frequency
+        dOm_tEds = vth*dM_tds/R0
```
krystophny, line 181: test

PR #41, opened 2026-05-28
Unify libneo resolver via LIBNEO_REF, drop $CODE shortcut

krystophny:
Adopts the canonical find_or_fetch shared across the codes: ref precedence is `<DEP>_REF` (cache or env, a branch/tag/commit validated against the remote and ignored if absent), then `<DEP>_GIT_TAG`, then the committed `<DEP>_RELEASE`, then the matching branch or main. Drops the legacy `$CODE` local-path shortcut; dependencies are always fetched at the resolved ref.

NEO-RT consumes libneo transitively through NEO-2. Companion CI change adds the `workflow_dispatch`/`libneo_ref` trigger.

### DIFF
--- a/cmake/Util.cmake
+++ b/cmake/Util.cmake
@@ -17,7 +17,16 @@ endfunction()
 
 function(fetch DEPENDENCY SOURCE_DIR)
     set(REPO_URL https://github.com/itpplasma/${DEPENDENCY}.git)
-    get_branch_or_main(${REPO_URL} REMOTE_BRANCH)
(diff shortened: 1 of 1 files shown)


krystophny:
@krystophny heads-up: this PR is part of a cross-repo change unifying how the codes resolve libneo (a validated `LIBNEO_REF` override and tracking the `release/26.2` branch by default), plus an external CI trigger so the libneo release gate can build this code against a candidate ref. It needed a few more changes than expected, so a review before it lands would help. We plan to merge the whole set on Monday, 2026-06-15. Companion CI PR: #42.

PR #46, opened 2026-06-14
Move non-event orbit integrations onto the fortnum ODE solver

krystophny:
Move the non-event orbit integrations (`bounce_fast`) from DVODE_F90 onto the
fortnum nonstiff variable-order Adams integrator (`fortnum_ode_vode`, DVODE
MF=10 equivalent): single span `[0, taub]`, scalar relerr 1e-9, per-component
abserr 1e-10 (ITOL=2). Re-pins the fortnum dependency from v0.1.0 to fortnum
main 974dcf1.

## Call sites
- `src/orbit.f90` `bounce_fast`: replaces the `dvode_f90` MF=10 call with
  `vode_init` + `vode_integrate_to` over the single bounce span; maps the
  fortnum status onto the legacy `istate` convention (2 success, -1 step
  budget, 0 failure).
- `CMakeLists.txt`: `FetchContent` `GIT_TAG` -> `974dcf12e2599ffa5882380207d753c8e8e0a9f6`.

## Verification

Golden record `test/golden_record/test_golden_record.py` against the
DVODE_F90-generated `golden.h5` (bar rtol=1e-8, atol=1e-15), CONFIG=Fast,
worst relative deviation per output:

The `bounce_fast` path on fortnum vode agrees with the DVODE reference far
better than the prior DOP853 substitution (which was ~1e-4 to ~15% on
`output`):

| case  | output rel | output abs |
|-------|-----------|-----------|
| 0p100 | 2.9e-9    | 7.5e-13   |
| 0p200 | 5.3e-9    | 5.6e-12   |
| 0p300 | 9.9e-9    | 3.7e-12   |
| 0p400 | 9.3e-9    | 1.2e-12   |
| 0p500 | 2.5e-7    | 3.1e-12   |
| 0p600 | 3.6e-7    | 1.1e-11   |
| 0p700 | 1.7e-7    | 2.9e-10   |
| 0p800 | 5.8e-8    | 1.5e-10   |
| 0p900 | 4.7e-8    | 2.2e-10   |

`magfie` matches exactly (rel 0). The bounce-averaged `output` (D11/D12)
agrees to 1e-9..3.5e-7.

The golden test still reports 9 failed on this branch, for two reasons:
1. The remaining `integral` / `torque_integral` failures (rel ~1e-4) come from
   the event-path `bounce_integral`, which on this branch still uses the
   DOP853 chunked scan; its taub is not the DVODE root. That path is migrated
   in the stacked PR #47.
2. Even the `bounce_fast` agreement tops out at ~3.6e-7 (worst case 0p600),
   above the 1e-8 bar. fortnum vode matches DVODE_F90 to about its requested
   tolerance, not bit-for-bit (the fortnum suite asserts accuracy versus the
   analytic solution, not bit-identity with DVODE_F90), so the
   DVODE-generated golden cannot be reproduced at rtol=1e-8.

The golden record is regenerated and RK4-validated in the stacked #47, where the event-path `bounce_integral` moves onto fortnum and its taub stops latching onto multiple-period bounce returns at the near-separatrix trapped spline nodes (see `test/golden_record/REGENERATION.md`). This branch keeps the DVODE/DOP853 event path, so the regenerated golden lands with #47, not here. The `bounce_fast` change on this branch shifts only the bounce-averaged Om_tB/D11/D12 at the ~1e-7 level; an independent 2e6-substep RK4 of the orbit ODE confirms the fortnum Om_tB bounce average to <1e-6 across the trapped grid. No tolerances were weakened.

Note: fortnum pin updated to current main (92de6e9) after a fortnum history rewrite; old shas no longer resolve.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -66,7 +66,7 @@ include(FetchContent)
 if(NOT TARGET fortnum)
     FetchContent_Declare(
         fortnum
-        GIT_REPOSITORY git@github.com:lazy-fortran/fortnum.git
+        GIT_REPOSITORY https://github.com/lazy-fortran/fortnum.git
         GIT_TAG a7faa3c
     )
     FetchContent_MakeAvailable(fortnum)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -67,7 +67,7 @@ if(NOT TARGET fortnum)
     FetchContent_Declare(
         fortnum
         GIT_REPOSITORY https://github.com/lazy-fortran/fortnum.git
-        GIT_TAG 974dcf12e2599ffa5882380207d753c8e8e0a9f6
+        GIT_TAG 92de6e949a772cfffc73bb5295fe5e2b056b9c18
     )
     FetchContent_MakeAvailable(fortnum)
 endif()
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -67,7 +67,7 @@ if(NOT TARGET fortnum)
     FetchContent_Declare(
         fortnum
         GIT_REPOSITORY https://github.com/lazy-fortran/fortnum.git
-        GIT_TAG e7df01236d46f30fffd01135b1ba00a465b64dae
+        GIT_TAG 974dcf12e2599ffa5882380207d753c8e8e0a9f6
     )
     FetchContent_MakeAvailable(fortnum)
 endif()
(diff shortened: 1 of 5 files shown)


krystophny:
Superseded by #47, which landed the full fortnum migration (fetch + non-event + event search + drop vode) as one squash commit reconciled onto current main. The commits from this PR are included there.

PR #58, opened 2026-06-24
POTATO: complete the resonance sweep via frequency-form search (parallel, gated)

krystophny:
Makes the POTATO finite-orbit resonance sweep complete, then fast and parallel,
on top of current main (fortnum ODE solver, edge bananas #54, libneo link). The
rho_pol=0.99 #30835 sweep that previously did not finish now runs in ~148 s on 16
cores and covers the full radial and velocity range.

## Algorithm: search in canonical frequencies, not bounce phase

This is the behavioral change. The resonance condition was rooted as the bounce
phase displacement `Delta_phi_b + 2*pi*m/n = 0`. Both `Delta_phi_b` and the
bounce time `tau_b` diverge at the trapped-passing (X-point) separatrix, so the
polynomial interpolation grid (`sample_matrix`) refined without end and the
rho_pol=0.99 sweep did not finish.

The same condition is now rooted in the canonical frequencies
`m*Omega_b + n*Omega_phi = 0`, with `Omega_b = 2*pi/tau_b` and
`Omega_phi = Delta_phi_b/tau_b`. These stay finite at the separatrix. The roots
are mathematically the same (multiply by `tau_b/2*pi`), so lines that both forms
resolve are unchanged; the grid now converges, so the sweep completes and
resolves the near-separatrix lines the displacement form left uncomputed.

Supporting near-separatrix handling: exclude grazer sentinels (`Omega_b=0`,
non-closing orbits) from the root grid, clip each X-point class to its closing
boundary, floor the per-row refinement scale, and cap the node count at 5000.

## Speed

- Memoize `interpolate_class_doublecount` across the 45 modes (x-keyed, 95.9% hit
  rate).
- OpenMP over the grid-build node fills and the per-mode root search.
- Batch the log flushes to cut per-line fsync.

Two OpenMP fixes the determinism gate caught: the per-mode loop must not make
`toten`/`perpinv` threadprivate (`pertham` writes them only to hand the orbit
energy to `velo_res`), so they are decoupled into `toten_orb`/`perpinv_orb`; and
a `get_rescond` by-product (`psiast`, `taub`, `delphi`) passed through the
gfortran internal-procedure trampoline does not honor an OpenMP `private` clause,
so it is threadprivate module state. The `bmod_pert` lazy load uses
double-checked locking.

## Scaling

Floating-point bound, so one thread per physical core. On a 16C/32T host the
sweep scales 620, 290, 191, 148 s for 1, 4, 8, 16 threads (4.2x). All 32 logical
threads run ~187 s idle and 400 to 600 s under load, because the SMT siblings
oversubscribe the cores. `run_potato.sh` sets the thread count and binding;
`README.md` documents it.

## Behavior and scope

Versus current main the reslines change: the sweep now finishes and covers the
near-separatrix region that did not compute before. Reslines are bit-identical
across thread counts (the OpenMP determinism guarantee), not versus the old
displacement form. The new single-orbit trace (`itest_type=4`) and frequency
radial scan (`itest_type=5`) are opt-in; edge extension across the separatrix
defaults off.

Known residual: `Omega_b ~ A/(-ln d)` near the separatrix is bounded but has
infinite slope, so the steepest closing classes hit the node cap and a few
high-m edge lines are truncated. The follow-up is a log-factored spline (spline
`Omega_b*(-ln d)`, reconstruct), which also removes NEO-RT #57's tp-boundary
spline ringing. Not in this PR.

## Verification

Public-safe circular-equilibrium golden gate (no AUG data), per-energy-slice
reslines counts within 2% of the committed reference:

```
$ ctest --test-dir build -R resonance_golden --output-on-failure
    Start 1: resonance_golden
1/1 Test #1: resonance_golden .................   Passed   39.78 sec
100% tests passed, 0 tests failed out of 1
```

The buggy intermediate (privatized `get_rescond` by-products) yields per-slice
counts ~0.6x the golden, so the 2% gate separates correct from corrupt.


krystophny:
Closing: the frequency-form resonance sweep is superseded by the merged resonant-torque engine fix (#67) and the displacement-based root finder now on main. Preserving the work as archive/feature/potato-resonances.

Issue #60, opened 2026-06-25
POTATO: energy-slice OpenMP parallelism hangs at >1 thread (near-separatrix orbit non-convergence)


krystophny:
## Summary

Energy-slice OpenMP parallelism of the POTATO displacement resonance engine
(`itest_type=3`, `resonant_torque`) hangs at `OMP_NUM_THREADS > 1`. The work is
on branch `feat/potato-omp-energy-wip`.

## What works

- `OMP=1` produces output byte-identical to the serial binary.
- At `>1` thread, the resonance lines that are written are sorted-bit-identical
  to serial and deterministic across runs. The parallelization is numerically
  sound where it completes.

## The hang

At `>1` thread a worker integrates a near-separatrix orbit whose adaptive step
collapses to `~1e-320` and never progresses (`SRC/velo.f90`, `hpstar -> 0`, deep
in `odeint`). Other threads then wait at the parallel-region barrier forever. The
same orbit completes at `OMP=1`, so either a per-slice state is inconsistent on a
worker, or the displacement form's lack of a bounce-time cap is exposed by the
parallel sampling timing.

## Work already done on the branch

Per-slice global state made `threadprivate` or passed by argument:

- `global_invariants` (`toten`, `perpinv`, `sigma`, `dtau`).
- the two `sample_matrix` grid modules and the resonance bookkeeping in
  `resint_mod`.
- `find_all_roots_mod` search state; `odeint_mod` integrator scratch.
- `bmod_pert` lazy spline load guarded with double-checked locking.
- the orbit-region set (`allowed_region`, a derived type with allocatable
  components) moved out of `bounds_fixpoints_mod` into a passed `region_set_t`.
  This removed an `OMP=1` segfault caused by passing a bare allocatable to an
  external procedure without an explicit interface.

## Suspected remaining causes / next steps

1. Another per-slice mutable module variable read by the orbit RHS (`SRC/velo.f90`)
   or the field evaluation (beyond the `threadprivate` `psif`/`ierrfield`) that is
   not yet isolated.
2. The displacement form has no grazer / bounce-time cap (`fb_max_tau`, skip of
   non-integrable near-separatrix orbits) that the frequency form carries; the
   near-separatrix orbit may be genuinely non-convergent and only avoided
   serially by timing.
3. The fix that addresses the root cause is likely the fortnum integrator
   (variable-order, with convergence caps) the frequency form already uses. It
   changes results, so it needs its own validation.

## Reproduce

[...]

PR #61, opened 2026-06-25
POTATO: coarse-grain energy-slice OpenMP parallelism (WIP, blocked on #60)

krystophny:
**Draft / work in progress.** Blocked on #60.

## What

Coarse-grain OpenMP parallelism for the POTATO displacement resonance engine:
parallelize the independent total-energy slices of `resonant_torque` instead of
the per-class inner loops of #59. Energy slices are independent and roughly equal
cost, so this scales far better than the per-class fine-grain (#59 plateaus at
~1.85x from grid-build load imbalance).

Per-slice state is made `threadprivate` (`global_invariants`, the two
`sample_matrix` grid modules, the `resint_mod` bookkeeping, `find_all_roots`
search state, `odeint_mod` integrator scratch), `bmod_pert`'s lazy load is
locked, and the orbit-region set is refactored out of module state into a passed
`region_set_t` object. The energy loop runs under `!$omp parallel do` with a
`reduction` on the torque sums; the inner per-class regions stay serial via
`omp_set_max_active_levels(1)`.

## Status

- `OMP=1` is byte-identical to serial.
- At `>1` thread the reslines that are written are sorted-bit-identical to serial
  and deterministic across runs.
- It hangs at `>1` thread on a near-separatrix orbit whose step collapses
  (`velo.f90`, `hpstar -> 0`). Tracked in #60 with the suspected causes and the
  path forward.

## Verification

Small case (`nenerg=4`, `m=+-8`, `nbox=50`):

```
OMP=1  rc=0   reslines sorted-md5 == serial binary (byte-identical)
OMP>1  reslines sorted-md5 == OMP=1, stable across runs (deterministic)
OMP>1  hangs after writing the reslines (see #60)
```

History on the branch shows the two earlier dead ends (all_regions threadprivate
-> hang; bare allocatable dummy -> OMP=1 segfault) and the fixes.

## Relationship to #59

#59 (fine-grain, ~1.85x, validated) is the shippable parallelization today. This
branch supersedes it once #60 is resolved.

### DIFF
--- a/POTATO/CMakeLists.txt
+++ b/POTATO/CMakeLists.txt
@@ -12,6 +12,11 @@ include(Util)
 find_package(BLAS REQUIRED)
 find_package(LAPACK REQUIRED)
 find_package(HDF5 REQUIRED COMPONENTS Fortran)
+# OpenMP parallelizes the find_bounce-heavy resonance grid build
+# (SRC/sample_matrix.f90) and the per-mode resonance root search
+# (SRC/resonant_int.f90).  Linked to potato_base so -fopenmp reaches POTATO's own
+# sources and the runtime is linked.
+find_package(OpenMP REQUIRED)
 
 include(FetchContent)
 FetchContent_Declare(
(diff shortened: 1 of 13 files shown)


krystophny:
Superseded by the energy-slice POTATO parallelism and follow-up fixes already merged on main.

PR #74, opened 2026-07-18
Expose POTATO invariant handoff

### LINKED ISSUE #73: Expose POTATO invariant handoff contract
## Goal

Expose POTATO's native guiding-center invariants and normalization metadata through a small, documented handoff contract so a POTATO orbit or scan point can be reproduced in another guiding-center code without guessing a radial label or private normalization.

SIMPLE PR [itpplasma/SIMPLE#502](https://github.com/itpplasma/SIMPLE/pull/502) adds the consuming side: an axisymmetric preprocessor from `(H0, J_perp, p_phi)` to SIMPLE's existing `[s, theta, phi, p, lambda]` initial state. This issue is only the POTATO/NEO-RT producer side. It should not duplicate SIMPLE's inverse solver or introduce a SIMPLE dependency.

## Motivation

The current benchmark can match trajectories through physical `R,Z`, and the chart-map/VMEC surface conventions are documented. That is sufficient for the completed Rung 0 checks, but it is not a clean invariant-space interface for finite-orbit-width scans, especially near the axis where several regular section intersections and both inner/outer topologies can occur.

POTATO already works with the relevant invariant construction internally. Downstream code needs the exact values and conventions, rather than reconstructing them from formatted diagnostics or assuming that `rho_pol`, `rho_tor`, a section radius, and canonical momentum are interchangeable.

## Proposed public record

For a single-orbit result and for every frequency-scan row, expose a machine-readable record containing at least:

- `H0` in POTATO's public reference-velocity normalization;
- the perpendicular invariant used by the orbit construction (`J_perp`, with its exact definition and units documented);
- raw `psi_star = (c/q) P_phi` in the equilibrium/g-EQDSK flux gauge;
- `psi_axis` and `psi_edge` in the same units, gauge, and sign convention as `psi_star`;
- POTATO reference velocity `v0` and enough species/reference-energy metadata to compare it with another code;
- the section point `(R_c, phi, Z_c)` and `sigma = sign(v_parallel)`;
- POTATO's section/function branch identifier and orbit class where available;
- status/error code, including failed scan points rather than silently dropping them;
- physical frequency units and the existing `taub` convention (`t_b = taub/v0`).

A small derived outward-oriented value may also be provided for convenience,

```text
p_phi_out = sign(psi_edge - psi_axis) * (psi_star - psi_axis),
```

but the raw values must remain available so the transformation is auditable. The total canonical momentum contains the parallel term and therefore need not vanish on the magnetic axis.

Prefer a versioned NetCDF record or a documented stable text schema. A callable Fortran routine returning a plain derived type or flat arrays would also be useful, provided output is produced once per orbit/scan point rather than through a per-timestep language boundary.

## Convention requirements

- This is guiding-center state only; full gyro-orbit initialization is out of scope.
- Do not identify `rho_pol` numerically with SIMPLE/VMEC `rho_tor` or `s_tor`.
- Do not expose SIMPLE's private symplectic coordinate or its private `sqrt(2)` time scaling. POTATO exports its own physical/native quantities.
- State explicitly whether electrostatic potential energy is included in `H0` and the perpendicular invariant. The first SIMPLE consumer supports the magnetic Hamiltonian, so the initial cross-code acceptance case must use zero electrostatic potential.
- Keep HFS-to-LFS section ordering (`R_c`) separate from inner/outer orbit topology. “Inner” and “outer” denote whether the orbit encircles the central forbidden region; they are not synonyms for HFS/LFS.
- Preserve both parallel-sign branches and every regular section solution. No implicit preferred-root selection.

## Suggested implementation

1. Factor the invariant values already computed by POTATO into one side-effect-free producer routine.
2. Call that routine from both `itest_type = 4` and the frequency-scan path.
3. Add one versioned writer for the record above.
4. Document formulas, units, gauge, velocity normalization, time conversion, and topology identifiers next to the file/API contract.
5. Keep adapters in the benchmark repository until the record is stable; no Python interface is required for the first PR.

## Acceptance tests

- A zero-potential circular-tokamak single orbit exports finite `(H0, J_perp, psi_star, psi_axis, psi_edge, v0)` plus its section metadata.
- The scan output contains the same record for successful points and explicit status for failed points.
- Reconstructing SIMPLE starts with `states_from_potato_invariants(...)` from SIMPLE #502 preserves the three converted invariants within documented tolerances.
- At least one reconstructed state agrees with the exported POTATO section point in cylindrical coordinates within the equilibrium interpolation/root tolerance; all additional regular roots remain visible.
- Tests cover both signs of `psi_edge - psi_axis` and both signs of `v_parallel`.
- Tests use identical physical species/reference energy in both codes and separately verify the documented `v0_POTATO/v0_SIMPLE` scaling case.
- `taub/v0` agrees with the physical bounce time used to form `omega_b`; no `sqrt(2)` correction appears.
- Existing POTATO inputs, `freq_scan.dat`, and orbit integration behavior remain compatible unless a schema version explicitly opts into additional columns.

## Benchmark follow-up

Once the producer contract is available, update `benchmark-simple-potato` so invariant handoff is the primary matched-orbit initialization route. Retain the existing physical `R,Z` overlay as an independent validation, not as the definition of invariant equivalence.

The multiple-root and topology requirements follow the finite-orbit construction discussed in Buchholz et al., *J. Phys.: Conf. Ser.* **2397** (2022) 012012, DOI [10.1088/1742-6596/2397/1/012012](https://doi.org/10.1088/1742-6596/2397/1/012012).

krystophny:
## Summary
- add a side-effect-free POTATO invariant producer for `H0`, `J_perp`, and raw `psi_star`
- write a versioned `potato_invariants.dat` from single-orbit and frequency-scan modes without changing existing output schemas
- document cgs flux, reference-velocity, electrostatic-potential, time, and topology conventions
- add unit and single/scan integration coverage

Closes #73.

## Validation
- Release POTATO build succeeds
- all 6 configured POTATO tests pass (2 fixture-dependent tests skip as designed)
- `fo fmt --check` and `fo lint` pass
- repository-level `fo check` remains blocked by the pre-existing `neort_setup_at_s`/`neort_lib` mismatch outside POTATO
- benchmark sibling end-to-end test reconstructs the reactor-size orbit in SIMPLE and matches both frequencies within 2%

### DIFF
--- a/POTATO/CMakeLists.txt
+++ b/POTATO/CMakeLists.txt
@@ -96,6 +96,16 @@ add_test(NAME test_wall_loss
    COMMAND test_wall_loss.x
 )
 
+add_executable(test_potato_invariants.x
+   SRC/test_potato_invariants.f90
+)
+target_link_libraries(test_potato_invariants.x
+   potato_base
+)
+add_test(NAME test_potato_invariants
+   COMMAND test_potato_invariants.x
+)
+
 add_executable(test_wall_profile.x
    SRC/test_wall_profile.f90
 )
@@ -126,6 +136,13 @@ target_link_libraries(potato_resonance_probe.x
 
 find_package(Python3 COMPONENTS Interpreter)
 if(Python3_Interpreter_FOUND)
+    add_test(NAME potato_invariant_handoff
+        COMMAND ${Python3_EXECUTABLE}
+            ${CMAKE_CURRENT_SOURCE_DIR}/test/invariant_handoff/test_invariant_handoff.py
+            $<TARGET_FILE:potato.x>
+            ${CMAKE_CURRENT_SOURCE_DIR}/test/golden_record_resonance
+    )
+    set_tests_properties(potato_invariant_handoff PROPERTIES TIMEOUT 150)
     add_test(NAME potato_golden_displacement
         COMMAND ${Python3_EXECUTABLE}
             ${CMAKE_CURRENT_SOURCE_DIR}/test/golden_record_displacement/test_potato_golden.py
--- a/POTATO/README.md
+++ b/POTATO/README.md
@@ -43,3 +43,33 @@ on an idle machine and 400-600 s under any concurrent load, because they
 oversubscribe the 16 cores.
 
 The reslines output is identical for every thread count.
+
+## SIMPLE-compatible invariant handoff
+
+Single-orbit (`itest_type=4`) and radial-frequency-scan (`itest_type=5`) runs
+write `potato_invariants.dat` without changing the existing `fort.100` or
+`freq_scan.dat` schemas.  The versioned file contains the local state and the
+three axisymmetric invariants needed by SIMPLE:
+
+```text
+H0       = p^2 + phi_elec
+J_perp   = p^2 (1 - xi^2)/B
+psi_star = psi + rho0 p xi h_phi = (c/q) P_phi
+```
+
+Here `p=v/v0`, `xi=v_parallel/v`, `v0=sqrt(2 E_ref/m)` in cm/s, `B` is in
+POTATO's native gauss convention, `phi_elec=q Phi/E_ref`, and `psi_star`,
+`psi_axis`, and `psi_edge` use the same native g-eqdsk flux gauge and units
+(G cm^2).  Time remains POTATO's `tau=v0*t`; SIMPLE's private symplectic
+sqrt(2) scaling is not part of this file.
+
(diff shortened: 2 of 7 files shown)


krystophny:
Both Actions jobs stop at the repository-wide `test/test_frequencies.f90` compile error (`neort_setup_at_s` missing from `neort_lib`), before POTATO tests run. This reproduces locally with `fo check` and is also present on the latest `main` runs, e.g. https://github.com/itpplasma/NEO-RT/actions/runs/29577767287. The focused Release POTATO build and all configured POTATO tests are green.

PR #78, opened 2026-07-20
Add optional collisional boundary-layer factor on resonance quadrature

krystophny:
## Scope

Adds an optional collisional boundary-layer factor to the resonance quadrature.

When `collisional_layer=.true.`, the transport and torque contributions are multiplied by a factor based on the local deflection frequency, resonance distance, and trapped/passing boundaries. The implementation includes the tabulated/interpolated layer ratio, elliptic-integral helpers, and the required configuration plumbing.

The default is `.false.`; with the option absent or disabled, the existing collisionless accumulation path is preserved. This PR does not alter the independent perturbation-input switch or other current-main configuration behavior.

## Validation

- Elliptic-integral, table-node, clamping, far-limit, collisionless-limit, and high-collisionality unit checks.
- Ripple-plateau integration tests verify default-off byte identity, the collisionless limit, and high-collisionality suppression.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -98,6 +98,7 @@ add_executable(${PROJECT_EXE_NAME} src/program.f90)
 set(NEO_RT_SOURCES
     src/attenuation_factor.f90
     src/collis_nbi.f90
+    src/collisional_layer.f90
     src/config.f90
     src/datatypes.f90
     src/driftorbit.f90
--- a/src/collisional_layer.f90
+++ b/src/collisional_layer.f90
@@ -1,0 +1,265 @@
+! Collisional boundary-layer factor for the resonance quadrature.
+!
+! Optional closure that multiplies each resonance-root contribution of the
+! collisionless (delta-function) quadrature in compute_transport_integral by a
+! factor R in [0, ~1.7] that accounts for the finite collisional width of the
+! resonance layer and its proximity to the trapped-domain boundaries.
+!
+! Derivation and verification: CAS-verified two-boundary Airy layer,
+! project evidence repo review/ell0_layer/neort_layer_kernel.wl
+! (7/7 CAS checks, 2026-07-20).  The scaled two-boundary layer equation is
+!   i z G - G'' = 1 ,  reflecting (Neumann) deeply-trapped boundary at z=-d_dt,
+!   absorbing (Dirichlet) trapped-passing seam at z=+d_tp,  R = Re<int G dz>/pi.
+! R is tabulated versus the two boundary distances scaled by the layer width
+! (R_TABLE below) and interpolated at run time.
+!
+! The NTVTOK detrapping enhancement nu_d/(2 eps) is NOT inserted by hand; it
+! emerges from the eta-space geometry through the bounce-averaged pitch
+! diffusion D_eta_ba and the relative layer width (see .wl section 1).
+!
+! Enable with the namelist logical collisional_layer = .true. (default off is
+! bit-identical to the collisionless quadrature).
+module neort_collisional_layer
+    use iso_fortran_env, only: dp => real64
+    use util, only: pi
+
+    implicit none
+    private
+
+    public :: collisional_layer_factor
+    public :: layer_ratio, layer_delta_eta, deflection_frequency
+    public :: comp_ellint_k, comp_ellint_e
+
+    ! Namelist switch (shared read-only configuration, mirrors driftorbit::supban).
+    logical, public :: collisional_layer = .false.
+
+    ! Electron mass (g).  util.f90 keeps only ion-scale constants, so the
+    ! electron thermal velocity needs its own documented constant here.
+    real(dp), parameter :: me = 9.109382e-28_dp
+
+    ! --- Two-boundary layer ratio table R(d_dt/delta, d_tp/delta) ------------
+    ! Generated from review/ell0_layer/layer_ratio_table.csv, itself produced by
+    ! review/ell0_layer/neort_layer_kernel.wl (LinearSolve of the scaled layer
+    ! equation on an 8x8 grid of scaled boundary distances 2^{-1..6}).
+    ! First index: d_dt/delta (deeply-trapped, reflecting side).
+    ! Second index: d_tp/delta (trapped-passing, absorbing side).
+    integer, parameter :: NGRID = 8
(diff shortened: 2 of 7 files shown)


krystophny (commented):
Reviewed latest `2037395f` after merging current main. The optional collisional layer remains default-off and its absent/explicit-off bit-identity test is now clearer. The review claim that `9.109382e-28` had fewer digits than `9.1094e-28` was arithmetically false, so the more precise existing constant was correctly retained. Full `fo` passes. No blocking finding.

krystophny:
Rebased onto current `main` after the converter/profile merges. This remains an optional collisional-layer feature, separate from the resonance-bracket, POTATO, converter, and full-orbit cleanup PRs. The existing local validation covered the focused CTest and ripple-plateau pytest gate; review still needs to resolve the outstanding requested changes.

krystophny (commented):
**Review verdict: Comment**

**Summary:**
This PR adds an optional collisional boundary-layer factor (`collisional_layer`, default `.false.`) that multiplies each resonance-root contribution in the transport quadrature by a tabulated ratio R accounting for the finite collisional width of the resonance layer and its proximity to the trapped-domain boundaries. It adds `src/collisional_layer.f90` (R-table interpolation, bounce-averaged pitch-diffusion layer width, deflection frequency), wires it through the namelist in `src/config.f90` and the accumulation in `src/transport.f90`, and adds a Fortran unit test plus Python integration tests. I verified the new module compiles, the unit test passes (`test_collisional_layer ... OK`), the Python tests are syntactically valid, and the default-off path is byte-identical (no multiplication inserted). The code degrades gracefully (R=1) when collisionality/profiles are unavailable and matches existing repo patterns.

**Findings:**
1. [minor] `src/collisional_layer.f90:16` — `collisional_layer` is a public mutable module global, contradicting the repo CLAUDE.md "NO GLOBAL MUTABLE STATE" mandate. It mirrors the existing `driftorbit::supban`/`magdrift` pattern, so it is consistent with current practice, but routing it through a read-only config argument would be cleaner.
2. [minor] `src/collisional_layer.f90:44-56` — the far-far table corner `R(64,64) = 1.000437` is not exactly 1.0, so an interior root "many layer widths" from both boundaries carries a systematic ~0.04% enhancement rather than the stated collisionless R=1 limit. Unit-test tolerances (2–3%) mask this; normalizing the table so the far corner equals 1 would be more faithful to the physics.
3. [minor] `src/collisional_layer.f90:189-192` — when plasma profiles are not loaded (`ni1≤0`, `Te≤0`, or `vth≤0`), `nu_d=0`, `delta_eta=0`, and the factor silently returns R=1 with no warning. Enabling the flag on a torque-free deck silently does nothing; a warning or doc note would help.
4. [minor] The new namelist is not documented in `examples/base/driftorbit.in` or `doc/`, and the table/derivation is referenced only as an external CAS notebook (`review/ell0_layer/neort_layer_kernel.wl`, dated 2026-07-20) absent from the repo, so the R-table's physical correctness cannot be independently verified from the PR alone. The interpolation/consistency unit tests are solid; the physics validation is external.
5. [minor] `test/ripple_plateau/test_collisional_layer.py` — the two physics tests are `fast_only` (skipped in default Debug runs) and all integration tests `pytest.skip` if `neo_rt.x` is not prebuilt. The unconditional CTest unit test covers the core logic.

**Verdict:** Comment — the change is clean, default-off safe, and well-tested; the findings are non-blocking polish items (global-state style, far-corner normalization, silent no-op behavior, and external physics validation), none of which require rework to merge.

krystophny:

## Summary
PR #78 adds an optional collisional boundary-layer factor (`collisional_layer`) that multiplies each resonance-root contribution of the collisionless delta-function transport quadrature by an interpolated two-boundary Airy-layer ratio `R`. The change is cleanly scoped: a new module `src/collisional_layer.f90`, a namelist switch wired through `config.f90`, guarded multiplication in `compute_transport_integral` (byte-identical when off, verified by a test), plus Fortran unit tests and Python integration tests. The Fortran is idiomatic, `pure`/`elemental` helpers are used, OpenMP threadprivate inputs are read consistently, and the default-off path is provably bit-identical. The main shortcoming is that the numerical heart of the feature — the `R_TABLE` — has no in-repo derivation or independent verification.

## Findings
1. **major** `src/collisional_layer.f90:33-65` (`R_TABLE`) — The core physics values come entirely from an external artifact (`review/ell0_layer/neort_layer_kernel.wl` and `layer_ratio_table.csv`) that is *not* in this repository; only a dated comment references it. The unit tests (`test_table_nodes`, `test_table_clamping`, `test_table_far_limit`) are self-referential — they re-read the same table constants at 1e-9 tolerance, so they validate interpolation mechanics, not physics. The integration tests only check the two asymptotic limits (collisionless → 1, huge collisionality → suppressed below 0.6); the intermediate `R` values that actually change transport results are never checked against an independent computation. Required fix: commit the derivation notebook/CSV (or a reproducible generator) into the repo, and/or add a test against an independently computed reference.

2. **minor** `src/collisional_layer.f90:26` — A new module-level mutable global (`logical, public :: collisional_layer`) is introduced, which the repository's CLAUDE.md explicitly forbids ("NO GLOBAL MUTABLE STATE — all state must be explicitly passed as parameters"). It mirrors the existing `driftorbit::supban`/config pattern, so it is consistent with current code, but it perpetuates the anti-pattern the project's own mandatory rules target. Consider threading it as an argument.

3. **minor** `src/collisional_layer.f90:214-250` (`deflection_frequency`) — This function is recomputed for every resonance root at the same velocity inside the velocity loop of `compute_transport_integral` (redundant work), and the claimed equivalence to NTVTOK's `nu_d` plus the hard-coded electron mass / Chandrasekhar-series constants have no in-repo test (e.g., `me` is never exercised by a unit test). The `erf(x_b) − G(x_b)` form is plausible and dimensionally consistent, but it would benefit from a small regression test against known values. Note also the documented limitation that passing particles get `R = 1` (unmodelled passing/loss coupling) — acceptable if stated in the PR description.

## Verdict
Request changes — the feature is well integrated and tested for mechanics and asymptotic behavior, but the physics table that determines all intermediate results is an opaque, externally sourced constant array with no in-repo derivation or independent numerical check, which the repository's test-first/evidence rules require.

PR #80, opened 2026-07-21
Fail closed on incomplete POTATO torque accounting

### LINKED ISSUE #82: Zero-rotation resonance roots cannot satisfy the convergence predicate
## Summary

`driftorbit_root` cannot report convergence when its caller supplies a zero
absolute tolerance.  Production transport and diagnostic callers compute that
tolerance as

```fortran
1.0e-8_dp * abs(Om_tE)
```

so every exactly zero-rotation (`Om_tE == 0`) resonance root necessarily runs
to `maxit`, emits a non-convergence warning, and returns the last iterate and
derivative anyway.

This is a numerical-control-flow defect.  It is separate from the physical
choice of rotation sign.

## Source-level reproduction

At revision `e4b862a55452033f478a5df5e46989f1085a121b`:

1. `src/transport.f90:100` and the other callers pass
   `1.0e-8_dp * abs(Om_tE)`.
2. For `Om_tE == 0`, `tol == 0`.
3. `src/resonance.f90:118` accepts a root only when
   `abs(res) < tol`.
4. Since `abs(res) >= 0`, that predicate is false for every finite residual,
   including an exact zero.
5. After 100 bisections the function stores the final derivative, warns, and
   returns the final iterate without a failure status.  Transport then consumes
   it.

A minimal behavioral regression is an otherwise valid sign-change bracket
evaluated with `tol=0`: the solver must terminate under a documented
residual/bracket-width rule and expose success or failure unambiguously.  The
existing `test_resonance_bracket` covers only the no-sign-change sentinel and
does not exercise this case.

## Actual behavior

- Exactly zero rotation can never satisfy the convergence predicate.
- Very small rotation makes the absolute tolerance depend on an unrelated
  vanishing profile value rather than on a documented resonance-frequency or
  bracket scale.
- The returned value is still used despite the warning, because the return
  value has no convergence status.

## Expected behavior

- A valid sign-change bracket should converge for zero and small `Om_tE` using
  a documented mixed residual/bracket-width stopping rule with a nonzero
  problem scale.
- Failure should not be indistinguishable from a valid two-element
  `(eta_root, dres/deta)` result.
- A test with an independent analytic or manufactured root should cover
  `Om_tE == 0`, not merely assert source state.

## Environment

- NEO-RT revision: `e4b862a55452033f478a5df5e46989f1085a121b`
- GNU Fortran / optimized OpenMP build through the repository `fo` workflow
- Observed in a radial NTV diagnostic campaign, but the defect follows directly
  from the source and is not device-specific.

## Scope note

This issue does **not** establish which sign of a MARS rotation profile should
be supplied to NEO-RT, nor does it validate torque normalization.  It only
blocks using the zero-rotation lane as clean resonance evidence.

krystophny:
## Summary

Final fail-closed POTATO torque-accounting layer:

- make Poincare cuts and topology bounds equilibrium-relative
- propagate sampler, orbit, resonance, and box-disposition failures explicitly
- reject incomplete energy slices and nonpositive bounce times
- serialize and cross-check scalar, energy, canonical-mode, root, and radial-box torque sums
- expose sampler controls without silently loosening them

## Validation

The rebuilt cumulative stack passes the full `fo` pipeline. Corrected TC24 L0 closes every serialized sum; no L5 convergence claim is made.

Stack: #100 -> this PR. No cross-code sign, phase, gain, smoothing, normalization, or radial fit is selected.

### DIFF
--- a/POTATO/CMakeLists.txt
+++ b/POTATO/CMakeLists.txt
@@ -122,6 +122,16 @@ add_test(NAME test_potato_invariants
    COMMAND test_potato_invariants.x
 )
 
+add_executable(test_matrix_callback_status.x
+   SRC/test_matrix_callback_status.f90
+)
+target_link_libraries(test_matrix_callback_status.x
+   "$<LINK_GROUP:RESCAN,potato,potato_base>"
+)
+add_test(NAME test_matrix_callback_status
+   COMMAND test_matrix_callback_status.x
+)
+
 add_executable(test_wall_profile.x
    SRC/test_wall_profile.f90
 )
--- a/POTATO/CMakeLists.txt
+++ b/POTATO/CMakeLists.txt
@@ -188,6 +188,13 @@ target_link_libraries(potato_hamiltonian_probe.x
    "$<LINK_GROUP:RESCAN,potato,potato_base>"
 )
 
+add_executable(potato_poicut_probe.x
+   SRC/potato_poicut_probe.f90
+)
+target_link_libraries(potato_poicut_probe.x
+   "$<LINK_GROUP:RESCAN,potato,potato_base>"
+)
+
 find_package(Python3 COMPONENTS Interpreter)
 if(Python3_Interpreter_FOUND)
     add_test(NAME potato_bmod_converter
@@ -226,4 +233,11 @@ if(Python3_Interpreter_FOUND)
         SKIP_RETURN_CODE 77
         TIMEOUT 240
     )
+    add_test(NAME potato_poicut_translation
+        COMMAND ${Python3_EXECUTABLE}
+            ${CMAKE_CURRENT_SOURCE_DIR}/test/poicut_translation/test_poicut_translation.py
+            $<TARGET_FILE:potato_poicut_probe.x>
+            ${CMAKE_CURRENT_SOURCE_DIR}/test/golden_record_resonance
+    )
+    set_tests_properties(potato_poicut_translation PROPERTIES TIMEOUT 150)
 endif()
--- a/POTATO/CMakeLists.txt
+++ b/POTATO/CMakeLists.txt
@@ -132,6 +132,16 @@ add_test(NAME test_matrix_callback_status
    COMMAND test_matrix_callback_status.x
 )
 
+add_executable(test_potato_input_validation.x
+   SRC/test_potato_input_validation.f90
+)
+target_link_libraries(test_potato_input_validation.x
+   potato_base
+)
+add_test(NAME test_potato_input_validation
+   COMMAND test_potato_input_validation.x
+)
+
 add_executable(test_wall_profile.x
    SRC/test_wall_profile.f90
 )
--- a/POTATO/CMakeLists.txt
+++ b/POTATO/CMakeLists.txt
@@ -98,6 +98,16 @@ add_test(NAME test_signed_toroidal_bound
    COMMAND test_signed_toroidal_bound.x
 )
 
+add_executable(test_box_counting_status.x
+   SRC/test_box_counting_status.f90
+)
+target_link_libraries(test_box_counting_status.x
+   "$<LINK_GROUP:RESCAN,potato,potato_base>"
+)
+add_test(NAME test_box_counting_status
+   COMMAND test_box_counting_status.x
+)
+
 add_executable(test_wall_loss.x
    SRC/test_wall_loss.f90
 )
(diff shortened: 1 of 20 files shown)


krystophny:
Sign/helicity audit follow-up: bd5eac0 maps the Boozer single-n coefficient onto geometric R-Z with A_RZ=A_B exp[i n (phi_B-phi_geom)] and adds a manufactured poloidally varying toroidal-shift regression. Full `fo` passes. On the TC24 chartmap the correction is ~2.2e-4 rad at s_tor=0.04 (max ~0.105 rad by s_tor~0.704), so it closes an independent representation defect but cannot explain the core ell=0 resonance. Accepted r5 remains immutable; any corrected campaign will use a new RUN_TAG.

krystophny:
Corrected TC24 production validation is complete for `bd5eac0`.

- Core R-Z grid gate: 513² / 1025² / 2049² with 2048 reconstruction angles. At `s_tor=0.04`, 1025→2049 changes native torque by 0.0015% (`phiI000`) and 0.0046% (`phiI010`), so production used 1025².
- aCluster job `19748663`: exit 0:0 in 26:57, 175/175 surfaces per phase, all remote and local SHA-256 gates passed, collisionless, `vsteps=2048`, `vmax/vth=4`, native `n_CC=-3` and native torque signs.
- Corrected core torque: +0.0581449 / -0.0566057 N m m^-3. Relative to pre-fix r5, the full profiles change by only 3.33% / 2.16% in relative L2 and correlate 0.99961 / 0.99975. Thus the coordinate-phase fix is required, but it does not explain the separate representation or rotation-sign discrepancy.

Compact evidence and refreshed figure are in project commit `aec3a084`: https://gitlab.tugraz.at/plasma/proj/ntv/iter_tc24/-/commit/aec3a084

krystophny:
Linking the issues this branch already resolves, so they close on merge. Each was verified against the branch by an independent review that was specifically asked to find the fixes *insufficient*:

- Closes #82 — `abs(res) < 0` could never succeed and the last iterate was consumed silently. `src/resonance.f90` now carries a scaled endpoint residual floor, an explicit convergence status, and a negative non-convergence sentinel.
- Closes #86 — the `50` in the POTATO profile header is a row count, not a format marker. `POTATO/tools/neo_rt_profiles_to_potato.py` now accepts any integral count greater than one and verifies the supplied rows.
- Closes #87 — `find_poicut` is now axis-relative and fail-closed, and `rhopol_boundary` uses safeguarded bounded updates, backed by a translation-invariance regression.
- Closes #88 — failed orbit classes were silently replaced by zero. `POTATO/SRC/resonant_int.f90` now propagates status, serializes completeness, and rejects any incomplete slice that is nonzero. This is distinct from #87: orbit-class completeness against geometry construction.

Not closed by this branch, tracked separately:

- #85 is addressed by #91, which is stacked above this one, and that PR does not claim to close it.
- #83 and #84 are untouched here.

krystophny (commented):
Reviewed latest head `a2b3d70` after merging current `main`. I checked the conflict resolutions specifically: both `magdrift_passing` and `inp_swi_pert` remain independent configuration inputs, the runtime Boozer handedness factor from main is retained, and both the EQDSK-axis and Boozer-handedness behavioral tests remain registered. GitHub Fast/Debug CI is green, and the complete stacked `fo` pipeline also passes locally.

Review finding: this remains four logically independent layers across 51 files (~3.9k added lines): direct-field input, adapters, POTATO fail-closed accounting, and diagnostics/tests. That is too broad for a reliable single merge review and should be split (or at minimum converted into a documented, ordered stack) before merge. I found no merge-resolution regression, but this scope/readability issue is blocking for claiming the whole PR independently reviewed.

krystophny (commented):
Reviewed latest `bc351872` relative to the new #101 base. The prior reviewability blocker is resolved by #99/#100/#101; this PR now contains only the fail-closed POTATO accounting layer. Sampler/orbit/root/box failures propagate explicitly and serialized scalar, energy, mode, root, and radial sums cross-check. Integrated `fo` is green. No blocking finding.

krystophny:
The useful orbit-failure subset of this broader stack is now isolated in #123: explicit get_matrix failure status plus adaptive-grid validity tracking. The remaining fail-closed accounting layers in this PR are still deferred as too broad/stacked for the current-main bugfix wave.

krystophny:
Closing this cumulative accounting stack as superseded for the current small-PR wave. The orbit-failure/status behavior is isolated in #123; the remaining multi-layer POTATO torque accounting is deferred to the dedicated torque work and is not being smuggled into a bug-fix PR. The branch and discussion remain available as historical reference.

PR #90, opened 2026-07-24
Serialize opt-in native resonance torque ledgers

krystophny:
## Scope

Adds an opt-in resonance ledger for diagnosing and independently reconstructing native torque contributions.

When `NEORT_RESONANCE_LEDGER` names an output file and `comptorque=.true.`, NEO-RT appends one row per resonance root with the surface, `mth/mph`, orbit branch, frequencies, resonance residual, derivative, bounce time, perturbation weight, attenuation, torque-density contribution, and integrator status. The default path and native torque result are unchanged when the environment variable is absent.

The bounce-debug diagnostic also accepts optional mode and branch filters, reports toroidal frequency/resonance residuals, and uses the configured `vmax_over_vth`.

## Validation

- `test_resonance_ledger_reconstructs_native_harmonic_torque` reconstructs each native branch contribution from the ledger.
- Existing golden-record tests remain the behavioral oracle.
- This PR is diagnostic infrastructure only; it does not change the production torque formula.

### DIFF
--- a/src/diag/diag_bounce_debug.f90
+++ b/src/diag/diag_bounce_debug.f90
@@ -1,178 +1,209 @@
 module diag_bounce_debug
-  use iso_fortran_env, only: dp => real64
-  use neort, only: init, check_magfie, write_magfie_data_to_files, &
-                   set_to_passing_region, set_to_trapped_region, vsteps
-  use neort_config, only: read_and_set_config
-  use neort_main, only: runname
-  use neort_datatypes, only: magfie_data_t
-  use neort_profiles, only: read_and_init_profile_input, read_and_init_plasma_input, init_profiles, vth, Om_tE
-  use neort_freq, only: Om_th
-  use neort_transport, only: timestep_transport
-  use neort_orbit, only: nvar
-  use neort_resonance, only: driftorbit_coarse, driftorbit_root
-  use driftorbit, only: mth, mph, mi, pertfile, etamin, etamax, sign_vpar, &
-                        etatp, etadt, efac
-  use do_magfie_mod, only: R0, s, q, bfac, do_magfie_init
-  use do_magfie_pert_mod, only: do_magfie_pert_init
-  use fortnum_ode, only: ode_problem_t, ode_workspace_t, ode_solution_t
-  use fortnum_ode_dop853, only: ode_integrate_dop
-  use fortnum_status, only: fortnum_status_t, FORTNUM_OK
-  implicit none
+    use iso_fortran_env, only: dp => real64
+    use neort, only: init, check_magfie, write_magfie_data_to_files, &
+        set_to_passing_region, set_to_trapped_region, vsteps, &
+        vmax_over_vth
+    use neort_config, only: read_and_set_config
+    use neort_main, only: runname
+    use neort_datatypes, only: magfie_data_t
+    use neort_profiles, only: read_and_init_profile_input, read_and_init_plasma_input, init_profiles, vth, Om_tE
+    use neort_freq, only: Om_th, Om_ph
+    use neort_transport, only: timestep_transport
+    use neort_orbit, only: nvar
+    use neort_resonance, only: driftorbit_coarse, driftorbit_root
+    use driftorbit, only: mth, mph, pertfile, etamin, etamax, sign_vpar, &
+        etatp, etadt, efac
+    use do_magfie_mod, only: R0, s, q, bfac, do_magfie_init
+    use do_magfie_pert_mod, only: do_magfie_pert_init
+    use fortnum_ode, only: ode_problem_t, ode_workspace_t, ode_solution_t
+    use fortnum_ode_dop853, only: ode_integrate_dop
+    use fortnum_status, only: fortnum_status_t, FORTNUM_OK
+    implicit none
 
 contains
 
-  subroutine run_bounce_debug(arg_runname)
-    character(*), intent(in) :: arg_runname
-    logical :: file_exists
-    integer :: i, j, mmin, mmax, nm, u
-    real(dp) :: vminp, vmaxp, vmint, vmaxt
-    real(dp) :: du, ux, v
-    real(dp) :: roots(100,3), eta_res(2), eta
-    integer :: nroots, kr
-    real(dp) :: Omth, dOmthdv, dOmthdeta, taub
-    integer :: istate
-    integer :: istats(50)
-    real(dp) :: rstats(50)
-    type(magfie_data_t) :: magfie_data
+    subroutine run_bounce_debug(arg_runname)
+        character(*), intent(in) :: arg_runname
+        character(32) :: arg_mth, arg_branch
+        logical :: file_exists
+        integer :: j, mmin, mmax, nm, u, target_mth
+        logical :: filter_mth
+        real(dp) :: vminp, vmaxp, vmint, vmaxt
+        real(dp) :: du, ux, v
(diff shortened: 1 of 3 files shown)


krystophny (commented):
Reviewed latest head `e6ddccf` against its updated #80 base after merging current `main`. The environment-gated path preserves default behavior when unset, refuses torque ledgers when `comptorque` is false, records the native canonical residual without changing its sign, and closes the append stream. The integrated `fo` pipeline passes.

Non-blocking but important gap: the PR body cites archived TC24 reconstruction at ~1e-14, but there is no checked-in behavioral test that enables `NEORT_RESONANCE_LEDGER` and independently reconstructs the native torque from emitted rows. The code is suitable as an opt-in diagnostic, but that reconstruction should become a regression oracle before treating the schema as a durable production contract. No blocking code finding for the stated diagnostic-only scope.

krystophny (commented):
Reviewed latest `d97892a8` relative to updated #80. The review gap is fixed: `test_resonance_ledger_reconstructs_native_harmonic_torque` runs the executable with the opt-in ledger and independently reconstructs every native harmonic/class torque. Focused result: 1 passed; full integrated `fo` is green. Default output remains unchanged when the environment variable is absent. No blocking finding.

krystophny:
Cross-reference from the #121 split: the unregistered test_bounce_integrand.f90 probe was intentionally omitted from the production fixes. If that diagnostic is still needed, it belongs with this resonance-ledger/frequency evidence PR and should be registered with a reproducible fixture.

krystophny:
Rebased and extracted onto current `main`. The old PR base carried direct-EQDSK perturbation-phase changes from the closed cumulative stack; those were deliberately not brought along. The rebased branch contains only the opt-in resonance ledger, the bounce diagnostic, and its golden-record reconstruction test.

Validation on the rebased head: `neo_rt.x` and `neo_rt_diag.x` build successfully; `NEO_RT_EXE=... python3 -m pytest -q` passes 16/16. Default behavior remains unchanged when `NEORT_RESONANCE_LEDGER` is unset. Follow-up is production AUG/ITER ledger comparison, not direct-field implementation in this PR.

krystophny:
**Review verdict: Approve**

# Review: PR #90 — Serialize native resonance torque integrands + bounce_debug enhancements

## Summary

This PR adds an opt-in diagnostics ledger (NEORT_RESONANCE_LEDGER) that serializes every resonance root contribution in compute_transport_integral (src/transport.f90), extends the bounce_debug diagnostic with per-mth/branch filtering, Omph, and resonance residual columns (src/diag/diag_bounce_debug.f90), and adds a golden-record integration test that reconstructs the native per-harmonic torque integral from the ledger and asserts it matches the _torque_integral.out output. I verified the 15-column ledger layout against its header, the branch-label logic (etamax < etatp + sign_vpar) against the three call sites in neort.f90, the du factor between code and test, and the reconstruction math (attenuation is not double-counted: the ledger stores dT/du without attenuation and the test multiplies by attenuation). The design is consistent and the test is well-targeted. No blockers found; findings below are minor.

## Findings

1. **minor** src/transport.f90:98 — The env_status == -1 "path is too long" guard is dead code. get_environment_variable is called without a VALUE buffer, so gfortran never returns -1 (it returns -1 only when a supplied buffer is too small); unset variables return a positive status. The intended overflow guard can never trigger, and a ledger path in a nonexistent directory will abort with a raw runtime error from open(). Recommend removing the misleading check or adding err=/inquire on the open to fail cleanly.

2. **minor** test/golden_record/test_golden_record.py:251-308 — The reconstruction test can pass vacuously: it never asserts that any ledger rows were written or that the reconstructed dict is non-empty. If a regression caused the ledger to emit no roots while the native torque entries at the chosen mid-range s happened to be ~0, the assert_allclose against the empty dict default 0.0 would still pass. Add an assert that at least one row was reconstructed.

3. **minor** src/diag/diag_bounce_debug.f90:72 — Behavior change for existing diagnostic users: vmaxp switched from a hardcoded 3.0*vth to config-driven vmax_over_vth (default 4.0), silently broadening the scan range and changing _bounce_debug.txt output. Reasonable, but should be noted in docs/changelog.

4. **minor** src/transport.f90:107 / src/diag/diag_bounce_debug.f90:80 — Header naming mismatch: columns named dF_deta / dres_deta actually contain eta_res(2), the bracket/derivative quantity used as the integrand denominator, not literally the residual derivative. Rename or document to avoid misinterpretation.

5. **minor** src/transport.f90 & src/diag/diag_bounce_debug.f90 — The diff is dominated by wholesale 2-to-4-space re-indentation interleaved with the real changes, making the substantive edits hard to review. A process nit; separate whitespace commits would help.

## Verdict

Comment — The change is well-designed and I verified the ledger layout, branch labeling, du scaling, and reconstruction math are mutually consistent, with a meaningful integration test tying the ledger back to native output. However, I could not execute the test suite here (read-only review), and the test's vacuous-passing weakness plus the minor issues above warrant non-blocking follow-ups rather than an unqualified approval.

krystophny:
**Review verdict: Approve**

Summary: This PR adds an opt-in diagnostic ledger (`NEORT_RESONANCE_LEDGER`) that serializes one row per resonance root in `src/transport.f90`, adds a golden-record test reconstructing per-harmonic/per-branch native torque from that ledger, and extends the `bounce_debug` diagnostic with `mth`/branch filters and toroidal-frequency/resonance-residual reporting. With the env var absent (default), the production path is unchanged, and CI (Fast + Debug builds) passes.

Findings:
1. [minor] `src/transport.f90:119` - the ledger writes `dT/du`, and the test (`test/golden_record/test_golden_record.py:302`) reconstructs by re-multiplying a `du` it derives from `(vmax_over_vth - 1.0e-6)/vsteps`. This only holds because `neort.f90:261-262` hardcodes `vminp = 1.0e-6*vth` and `vmaxp = vmax_over_vth*vth`. The reconstruction is implicitly coupled to those bounds; any future change to them (or a supban-style differing trapped bracket) would silently break or mislead the test. Consider writing `du` (or `vmin/vth`, `vmax/vth`) into the ledger header so the test recomputes it robustly.
2. [minor] `src/transport.f90:111-118` - the orbit-branch label is inferred from `etamax < etatp` + `sign_vpar`, duplicating the logic in `set_to_passing_region`/`set_to_trapped_region` (`neort.f90:56-72`). Currently correct for all three branches, but it is fragile duplicated knowledge; deriving the label from the same routine that sets the region would satisfy DRY.
3. [minor] `src/diag/diag_bounce_debug.f90:172` - `all(trim(arg_branch) /= ['all ','cop ','ctr ','trap'])` works only via Fortran blank-padding of the left operand; a length-4 explicit comparison or select-case would be clearer. The change of `vmaxp` to `vmax_over_vth*vth` (replacing the hardcoded `3.0*vth`) is a good improvement.

The ledger test is well constructed: it compares two outputs from the same run (native `_torque_integral.out` vs. ledger), so the tight `rtol=2e-13, atol=1e-280` is deterministic and not cross-CPU flaky; the `(mth, branch)` keying and column mapping `cop->1, ctr->2, trap->3` match `neort.f90:449`, and the `dT/du * du` arithmetic checks out against the native `T += du*Tphi_int/|eta_res|*attenuation` accumulation. No functional blocker found.

Verdict: Approve - correct, default-path-unchanged, diagnostic-only infrastructure; CI passes; the remaining items are minor maintainability/DRY couplings that do not affect correctness.

Issue #85, opened 2026-07-23
Direct-EQDSK magnetic precession disagrees with straight-field-line backend


krystophny:
> **Revised after review.** The numerical comparison originally quoted here rested
> on a `.bc` straight-field-line reference that was later found to carry the wrong
> toroidal-flux sign, which also invalidated the MARS `KDRIFT/ZOMEGADT`
> cross-check built on it. Those rows are **withdrawn** below rather than deleted,
> so the withdrawal is auditable. The issue does not depend on them: the substantive
> claim is a code-level one about which angle the field-line label uses, and it can
> be read directly out of the source.
>
> Also corrected: the original text implied the disagreement was *expected* as a
> chart or gauge effect. That is wrong. A chart change adds a total derivative whose
> closed-orbit average vanishes only when the **full** label gradient is retained,
> including its radial term. A persistent bounce-averaged difference is therefore a
> defect, not a gauge artefact — which strengthens this issue rather than weakening
> it.

At clean revision `88ad3a1`, the direct-EQDSK (`inp_swi=11`) and legacy
straight-field-line (`inp_swi=9`) backends give nearly identical trapped bounce
frequencies but incompatible magnetic-precession frequencies for two
representations of the same TC24 equilibrium. This blocks direct-EQDSK NTV torque.

## The code-level defect, which stands on its own

`src/orbit.f90`, in the `inp_swi == 11` branch introduced by `9e7f065`, computes
the physical magnetic-drift vector in `(s_tor, phi, theta_geo)` and then forms the
canonical toroidal precession as

```fortran
Om_tB_v = drift_phi_v - q*drift_theta_v      ! comment: alpha = phi - q*theta
```

`theta_geo` is the **geometric** poloidal angle, not a straight-field-line or
Boozer angle. So `phi - q*theta_geo` with the global `q` is not constant along a
field line, and a precession derived from it is not the canonical one. Two things
are wrong with it:

1. **Wrong pitch.** Along a field line `dphi/dtheta = B^phi/B^theta`, so the label
   gradient carries the *local* pitch `hctrvr(2)/hctrvr(3)`, which reduces to `q`
   only when `theta` is already a straight-field-line angle.
2. **Missing radial term.** The straight-field-line-angle map depends on `s`, so
   its gradient has a radial component. The `inp_swi=9` branch carries exactly such
   a term (`shearterm = Bphcov * dqds`); the `inp_swi=11` branch has no analogue.

[...]

PR #91, opened 2026-07-25
Validate the direct-EQDSK local field-line pitch

krystophny:
## Summary

The direct core already projects drift with the local field-line pitch. This PR now contains the independent validation and fail-closed guard only:

- integrate the field line directly in cylindrical coordinates and compare its rotational transform with the theta-average of `B^phi/B^theta`
- verify the local pitch genuinely varies with geometric theta
- reject a vanishing `B^theta` before forming the local pitch

The superseded duplicate implementation commit was dropped during the stack rebuild.

## Validation

- rebuilt cumulative stack passes full `fo`
- independent field-line integration agrees at `s=0.15,0.35,0.60`

Stack: #90 -> this PR.

### DIFF
--- a/src/orbit.f90
+++ b/src/orbit.f90
@@ -450,6 +450,9 @@ contains
             ! Geometric theta is not a straight-field-line angle.  Project the
             ! drift across the local field-line label so parallel motion has
             ! identically zero canonical toroidal component.
+            if (abs(hctrvr(3)) <= tiny(hctrvr(3))) then
+                error stop "direct-EQDSK B^theta vanishes; local field-line pitch is undefined"
+            end if
             Om_tB_v = fieldline_label_component( &
                 drift_phi_v, drift_theta_v, hctrvr(2), hctrvr(3))
         else
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -63,6 +63,14 @@ add_test(NAME test_boozer_handedness
     WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}
 )
 
+# Independent oracle for the direct-EQDSK field-line label: the theta-average of
+# B^phi/B^theta must reproduce the rotational transform obtained by integrating
+# the actual field line in cylindrical coordinates.
+add_test(NAME test_eqdsk_local_pitch
+    COMMAND ${CMAKE_BINARY_DIR}/test_eqdsk_local_pitch.x
+    WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}
+)
+
 add_test(NAME test_profile_collision_species
     COMMAND ${CMAKE_BINARY_DIR}/test_profile_collision_species.x
     WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}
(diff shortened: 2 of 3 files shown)


krystophny:
**Update: added a test, and found the reason one could not be written before.**

### The substituted quantity is now validated

The change substitutes `hctrvr(2)/hctrvr(3)` for the global `q`, and nothing here
tested that the substituted quantity is actually the field-line pitch. Commit
`71c39fa` adds `test_eqdsk_local_pitch`, whose oracle is a field line integrated
in cylindrical coordinates from `field_eq` alone:

```
dR/dphi = R B_R/B_phi ,   dZ/dphi = R B_Z/B_phi
```

advanced until the geometric angle about the magnetic axis completes one circuit,
giving the rotational transform directly as `Delta_phi/(2 pi)`. It shares no code
with `do_magfie_eqdsk`'s coordinate transformation, so it is an independent
behavioural oracle rather than a restatement of the patch. The two agree to
**7e-4** at s = 0.15, 0.35, 0.60.

The test also asserts the local pitch genuinely varies with theta — without that,
an implementation that simply returned the global `q` would satisfy the averaged
comparison, and the test would not discriminate the change it exists to protect.

### Why the obvious test could not be written against `q`

Writing it surfaced a separate and larger defect. `geoflux` built **both** the q
profile **and the `s` label** from the GEQDSK's stored `qpsi` record, and on
`circ.eqdsk` that record disagrees with the rotational transform of the fixture's
own field by a factor of about **1.55**. A test written against the tabulated `q`
therefore fails for a reason having nothing to do with the field-line label — and
that is exactly why `test_eqdsk_axis` could only afford a permissive
factor-of-two bound on the same comparison.

Fixed upstream in [itpplasma/libneo#404](https://github.com/itpplasma/libneo/pull/404),
which computes `q(psi) = (F/(2 pi)) oint dl_p/(R |grad psi|)` from `psirz` and
`fpol`. **This PR's new test needs that libneo change to pass.**

I briefly suspected this PR's own substituted pitch was mis-normalised, since its
theta-average was low against `q` by a factor close to `R0 = 1.6`. That was
wrong: the metric is right and the tabulated `q` was wrong. The field-line
integration settles it, and this PR's quantity is correct.

### Still not sufficient, unchanged from the original description

The radial term is still missing. `alpha = phi - q(s) f(s,theta)` has

```
grad(alpha) = grad(phi) - q_local grad(theta) - (q' f + q f_s) grad(s)
```

and only the first two terms are implemented here. `inp_swi=9` carries an
analogue of the third (`shearterm = Bphcov*dqds`); `inp_swi=11` has none. So this
remains necessary and not sufficient, and #85 stays open after it merges.

krystophny (commented):
Reviewed latest head `85d6ef8` against updated #90 after merging current `main`. The change uses the local direct-EQDSK field-line pitch `B^phi/B^theta` where geometric theta is not a straight-field-line angle. The new regression is a genuine independent oracle: it integrates the cylindrical field directly, requires non-constant local pitch, and checks the theta average. The full integrated `fo` pipeline passes.

The prior minor hardening note still applies: production divides by `hctrvr(2)` without the near-zero guard used by the test. That is not a blocker for the validated off-axis tokamak domain, but the precondition should eventually be explicit if near-axis orbits are admitted. No blocking finding.

krystophny (commented):
Reviewed latest `cee794ec` relative to updated #90. The local field-line pitch remains validated against independent cylindrical field-line integration, and the prior denominator hardening note is fixed: production now fails closed before dividing when direct-EQDSK `B^theta` vanishes. Full integrated `fo` is green. No blocking finding.

krystophny:
Closing this cumulative direct-EQDSK validation stack as deferred. The independently reviewable resonance-range guard is now #103; the direct-field-line validation belongs with the later direct-EQDSK feature work and is not a small mainline fix.

PR #94, opened 2026-07-25
Add magdrift_passing so one run reproduces MARS-K's orbit-class drift scope

krystophny:
MARS-K has no passing magnetic-drift code path. Matching that scope needed two NEO-RT runs whose per-class columns were summed by hand; this does it in one.

`magdrift_passing` gates the passing branch of `Om_tB` (now exactly zero) and `init_canon_freq_passing_spline`. The namelist control is an integer with a negative *follow magdrift* default, so every existing deck stays bit-identical and no deck must name the new key.

**Draft: validation pending, do not merge yet.** The oracle is the two-run equivalence — trapped-with-drift plus passing-without, summed from the separate `Tco`/`Tctr`/`Tt` columns, must reproduce a single run with `magdrift=.true.` and `magdrift_passing=0`. That compares against an existing independent code path rather than restating the patch. A bespoke unit test was attempted and dropped: driving `Om_tB` directly needs more orbit-machinery setup than the check is worth, and the two-run comparison is the stronger oracle.

Full suite passes (14/14) with the default, as expected since the default changes nothing.

First of a planned set of code-mimicry flags (collisionality including **passing**, which #78 currently leaves at R=1, smoothing, spline variants), whose purpose is to let NEO-RT run *as* MARS-K on the MARS field and *as* PENTRC on the GPEC field, so residual differences are physics rather than configuration. Context: ITER TC24 NTV benchmark.


krystophny:
Superseded by the clean, independently behavior-tested #97. The original branch accidentally included the #80/#90/#91 stack, so I am closing this oversized version rather than asking reviewers to disentangle 54 unrelated files.

PR #103, opened 2026-07-27
Keep resonance searches inside the spline pitch range

krystophny:
## Scope

Keeps resonance searches strictly inside the validated spline pitch range.

The trapped-side bracket now uses the validated trapped spline margin, while the passing-side bracket uses the passing spline margin plus a machine-precision guard. This prevents roots from being handed to explicit bounce integration exactly at or beyond the trapped-passing spline boundary.

## Validation

Adds and registers `test_resonance_spline_bounds`. The test initializes the base case, obtains both trapped and passing regions, and verifies that their bounds remain on the spline-safe side of the separatrix.

### DIFF
--- a/src/driftorbit.f90
+++ b/src/driftorbit.f90
@@ -39,7 +39,13 @@ module driftorbit
 
     real(dp), parameter :: epst_spl = 1.0e-6_dp, epsp_spl = 1.0e-6_dp   ! dist to tpb for spline
     real(dp), parameter :: epsst_spl = 1.0e-3_dp, epssp_spl = 1.0e-3_dp ! dist to deep for spline
-    real(dp), parameter :: epst = 1.0e-8_dp, epsp = 1.0e-8_dp ! smallest eta distance to tp bound
+    ! Keep resonance searches strictly on the spline side of the
+    ! trapped-passing boundary. Trapped roots need a larger validated margin
+    ! to bound explicit bounce integration at the separatrix; passing roots
+    ! only need a representable guard above the spline boundary.
+    real(dp), parameter :: passing_bracket_guard = 128.0_dp*epsilon(1.0_dp)
+    real(dp), parameter :: epst = 1.0e-5_dp
+    real(dp), parameter :: epsp = epsp_spl + passing_bracket_guard
 
 
 
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -74,6 +74,10 @@ add_test(NAME test_resonance_bracket
     COMMAND ${CMAKE_BINARY_DIR}/test_resonance_bracket.x
     WORKING_DIRECTORY ${CMAKE_SOURCE_DIR}/examples/base
 )
+add_test(NAME test_resonance_spline_bounds
+    COMMAND ${CMAKE_BINARY_DIR}/test_resonance_spline_bounds.x
+    WORKING_DIRECTORY ${CMAKE_SOURCE_DIR}/examples/base
+)
 add_test(NAME test_perturbation_fourier_kernel
     COMMAND ${CMAKE_BINARY_DIR}/probe_neort_perturbation_fourier_kernel.x
     WORKING_DIRECTORY ${CMAKE_BINARY_DIR}
--- a/test/test_resonance_spline_bounds.f90
+++ b/test/test_resonance_spline_bounds.f90
@@ -1,0 +1,28 @@
+program test_resonance_spline_bounds
+    use iso_fortran_env, only: dp => real64
+    use logger, only: set_log_level
+    use neort_lib, only: neort_init, neort_prepare_splines, neort_setup_at_s
+    use neort, only: set_to_passing_region, set_to_trapped_region
+    use driftorbit, only: etatp, etadt, epsp_spl, epsst_spl, epst_spl
+
+    implicit none
+
+    real(dp) :: eta_min, eta_max
+
+    call set_log_level(-1)
+    call neort_init("driftorbit.in", "in_file")
+    call neort_prepare_splines("plasma.in", "profile.in")
+    call neort_setup_at_s(0.5_dp)
+
+    call set_to_trapped_region(eta_min, eta_max)
+    if (eta_min < (1.0_dp + epst_spl) * etatp) then
+        error stop "trapped lower bound reaches the spline separatrix"
+    end if
+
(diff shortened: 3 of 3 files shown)


krystophny:
Rebased cleanly onto current `main` after the converter/profile merges; now independent of the old direct-EQDSK stack. Focus remains the one-line resonance spline-range guard plus its regression test.

krystophny:
Audit correction: after making this branch independent of the former direct-EQDSK stack, the old description still claimed a regression test that actually belonged to #104. Added a dedicated `test_resonance_spline_bounds` behavioral oracle and registered it in CTest. It initializes the production fixture and verifies that the trapped and passing resonance-search brackets stay on the supported side of the spline separatrix. Local Release build and focused CTest pass on current `main`.

krystophny:

Summary:
PR #103 ("Keep resonance search inside spline pitch range") changes the trapped–passing boundary guards in `src/driftorbit.f90`. Previously `epst = epsp = 1e-8`; now `epst = 1e-5` and `epsp = epsp_spl + 128*epsilon(1.0)` ≈ 1e-6. Both the trapped lower bound (`eta_min = (1+epst)*etatp`) and the passing upper bound (`eta_max = (1-epsp)*etatp`) move inward so resonance searches stay strictly inside the spline-valid pitch domain instead of dipping into the analytic log-extrapolation zone near the separatrix. A new ctest (`test_resonance_spline_bounds`) pins the new bounds against the spline constants.

Findings:
1. [major] src/driftorbit.f90:47-48 — The change silently excludes physical resonances and alters NTV/torque output, with no validation. Trapped pitch in `(1+1e-8, 1+1e-5)*etatp` and passing pitch in `(1-1e-6, 1-1e-8)*etatp` (relative to the separatrix) are now never searched for resonances. The commit claims a "validated margin" but no validation is shown: there is no before/after torque comparison, no check that no real roots were dropped, and the golden record (`test/golden_record/`, which stores `Dt_trap Dt_pass Dt …`) is not regenerated or verified against the new results. The single test is a tautology that only checks the new constants land on the spline side (it would pass for any sufficiently large `epst`), so it cannot establish correctness. Required: demonstrate with a torque/physics comparison that the excluded band contains no physical resonances (or that the previous extrapolation produced spurious roots), and update/validate the golden reference.

2. [minor] src/driftorbit.f90:47 — `epst = 1.0e-5_dp` is an arbitrary magic number inconsistent with the derived `epsp = epsp_spl + guard`. `epsp` is correctly tied to `epsp_spl` (1e-6) plus a representable guard, but `epst` is hard-coded 10× the trapped spline offset `epst_spl` (1e-6) with no derivation or rationale for the factor. The asymmetry between the trapped and passing margins is unexplained. Suggest deriving `epst` from `epst_spl` symmetrically (e.g., `epst = epst_spl + guard` or a named factor), and documenting why the trapped margin is larger.

3. [minor] test/test_resonance_spline_bounds.f90:1-28 — Coverage is thin. It verifies only the trapped lower bound and passing upper bound; it never exercises `driftorbit_root`/resonance bracketing near the boundary, never runs the transport path, and does not assert the passing lower bound (`eta_min = epsp*etatp`) or trapped upper bound (`eta_max = (1-epst)*etadt`). To be a meaningful regression test it should reproduce the failure the fix addresses (e.g., a spurious/incorrect root found in the extrapolation zone with the old constants) rather than only re-asserting the new constants.

Verdict: Request changes — the guard change alters physics output near the trapped–passing boundary without validation or a golden-record update, and the shipped test does not demonstrate correctness; the arbitrary `epst` and thin coverage should be addressed alongside a results comparison.

krystophny:
**Review verdict: Approve**

## Summary
PR #103 changes the pitch-angle search margins in `src/driftorbit.f90` so the trapped/passing resonance brackets stay on the spline-evaluated side of the trapped–passing separatrix instead of dipping into the singular logarithmic extrapolation zone, and adds a regression test (`test_resonance_spline_bounds.f90`) pinning the new margins. `epst` goes from 1e-8 to 1e-5 and `epsp` from 1e-8 to `epsp_spl + 128·epsilon`. The logic is correct: `epst=1e-5 ≥ epst_spl=1e-6` places the trapped lower bound inside the trapped spline range, and `epsp ≈ 1e-6 ≥ epsp_spl` places the passing upper bound inside the passing spline range; the new test fails on the base constants (1e-8 < 1e-6) and passes on the new ones, so it is a valid RED→GREEN regression test. The test is auto-discovered by the `test_*.f90` glob and uses the same `examples/base` fixture as the existing `test_resonance_bracket`.

## Findings
1. [minor] src/driftorbit.f90:41-48 — The comment and commit message claim the searches stay "inside the spline pitch range", but only the two separatrix-side boundaries are actually constrained. The passing *lower* bound `epsp*etatp ≈ 1e-6·etatp` is far below the passing spline's lower boundary `epssp_spl*etatp = 1e-3·etatp`, and the trapped *upper* bound `(1-epst)*etadt` can exceed the trapped spline upper boundary for strongly-trapped configurations. These are not numerically hazardous (no log singularity away from the separatrix) and are pre-existing, but the test only pins the two singular-side bounds, so the guarantee is narrower than stated. Consider rewording the comment to say the searches stay on the spline side of the separatrix rather than the full pitch range.
2. [minor] src/driftorbit.f90:44 — `passing_bracket_guard = 128.0_dp*epsilon(1.0_dp)`: the factor 128 is a magic number and `epsilon(1.0_dp)` mixes a default-real argument with the `dp` literal; harmless as a guard but could be a named constant with an explicit comment. Likewise, `epst` is hard-coded (1e-5) rather than derived from `epst_spl` the way `epsp` derives from `epsp_spl`; the test guards against `epst < epst_spl`, so drift is caught, but tying it to `epst_spl` would be more maintainable.

## Verdict
Approve — the change is correct, low-risk, and accompanied by a meaningful regression test that fails on the old constants; findings are non-blocking. (Full build/test execution was not possible in this offline review environment because the build fetches spline/fortnum/fortplot/libneo via FetchContent and the `examples/` tree is absent from the checkout, so this review relies on static analysis.)

krystophny:
Superseded by PR #129. The fixed spline-margin approach in this PR is replaced by the validated generated separatrix fit, machine-adjacent logarithmic resonance scan, and independent geometry/oracle coverage now carried by #129. Closing this obsolete implementation PR; its history remains available for provenance.

PR #104, opened 2026-07-27
Make zero-rotation resonance roots converge deterministically

krystophny:
## Summary

Make zero-rotation resonance roots converge with a scale-aware numerical floor and return a defined failure sentinel when bisection cannot converge.

The public `driftorbit_root` signature remains unchanged. An optional convergence-status output was audited and removed because no production caller or other PR consumes it.

## Validation

- Rebased independently onto current `main`; #105 is no longer a parent.
- The compiled resonance regression locates a magnetic-drift root at zero electric rotation and independently checks its native residual.
- Focused Release validation passes.

No direct-EQDSK, converter, or full-orbit code is included.

### DIFF
--- a/src/resonance.f90
+++ b/src/resonance.f90
@@ -51,18 +51,24 @@ contains
         call driftorbit_coarse(v, eta_min, eta_max, roots, driftorbit_nroot)
     end function driftorbit_nroot
 
-    function driftorbit_root(v, tol, eta_min, eta_max)
+    function driftorbit_root(v, tol, eta_min, eta_max, converged)
         use logger, only: warning
 
+        ! Bisection can exhaust the representable eta values before a
+        ! zero-frequency lane reaches an absolute residual of exactly zero.
+        ! Keep the fallback well below the transport quadrature accuracy while
+        ! acknowledging the conditioning of the separatrix frequency splines.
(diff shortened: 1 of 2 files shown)


krystophny:
Rebased cleanly onto current `main` and made independent of #105. Focus remains deterministic zero-rotation resonance bracketing with its behavioral test.

krystophny:
Rebased onto current `main` and kept independent of #105. The change is limited to deterministic zero-rotation bisection termination and the existing resonance behavioral test; no converter, direct-field, or full-orbit code is included. Focused `test_resonance_bracket` validation passed during the rebase audit.

krystophny:
Manual workflow-dispatch validation completed successfully at rebased head 18839a2: build (Fast) and build (Debug) both passed. GitHub does not attach workflow-dispatch runs to the PR check list because this PR has no synchronize event after the history rewrite.

krystophny:
Audit result: the optional `converged` output had no consumer in production code, any open PR, or any planned follow-up. It existed only in this PR's test, so commit 938377c removes it.

The remaining change is limited to the live thin-orbit zero-rotation path: existing callers pass `1e-8*abs(Om_tE)`, which is zero at zero electric rotation. The regression now checks the returned root's native residual directly.

PR #102, opened 2026-07-27
Add validated POTATO input converters

krystophny:
## Summary

- convert signed Boozer `Delta|B|` spectra to POTATO R-Z perturbation grids without radial filtering
- preserve the Boozer-to-geometric toroidal phase relation explicitly
- convert native NEO-RT profiles from `s_tor` to POTATO `s_pol`, selecting the physical ion by charge and density
- serialize hashes, sign conventions, fit residuals, global gridding error, and per-surface gridding error
- document zero-field and mapped-domain behavior

## Validation

- full cumulative `fo` pipeline passes
- converter tests use analytic Fourier/phase and profile mappings plus binary round-trip checks

Stack: #106 -> #107 -> this PR. Extracted from the former mixed #99/#101 scope.

PR #105, opened 2026-07-27
Handle native signed modes in POTATO torque

krystophny:
## Scope

Corrects POTATO resonance handling for native signed toroidal modes and native flux orientation.

- The resonance search extent uses `abs(n)` only where a positive search magnitude is required.
- The signed `n` remains unchanged in the resonance equation and Fourier phase.
- SOL/LCFS classification is expressed relative to the axis-to-edge flux direction, so both increasing and decreasing EQDSK gauges behave correctly.
- The bound and orientation logic are isolated in `resonance_mode_bounds_mod`.

## Validation

Adds `test_signed_toroidal_bound`, covering positive and negative `n` and both flux orientations. The test compares the signed-mode bounds with a numerical tolerance and checks inside/outside classification.

### DIFF
--- a/POTATO/CMakeLists.txt
+++ b/POTATO/CMakeLists.txt
@@ -88,6 +88,16 @@ add_test(NAME test_find_all_roots_bracketed
    COMMAND test_find_all_roots_bracketed.x
 )
 
+add_executable(test_signed_toroidal_bound.x
+   SRC/test_signed_toroidal_bound.f90
+)
+target_link_libraries(test_signed_toroidal_bound.x
+   potato_base
+)
+add_test(NAME test_signed_toroidal_bound
+   COMMAND test_signed_toroidal_bound.x
+)
+
 add_executable(test_wall_loss.x
    SRC/test_wall_loss.f90
 )
--- a/POTATO/SRC/CMakeLists.txt
+++ b/POTATO/SRC/CMakeLists.txt
@@ -34,6 +34,7 @@ add_library(potato_base
   sorting.f90
   binsrc.f90
   bmod_pert.f90
+  resonance_mode_bounds_mod.f90
   resonant_int.f90
   eqmagprofs.f90
   box_counting.f90
--- a/POTATO/SRC/resonance_mode_bounds_mod.f90
+++ b/POTATO/SRC/resonance_mode_bounds_mod.f90
@@ -1,0 +1,30 @@
+module resonance_mode_bounds_mod
+    implicit none
+
+contains
+
+    pure function resonant_delphi_bound(m_modes, n_modes) result(bound)
+        integer, intent(in) :: m_modes(:), n_modes(:)
+        double precision :: bound
+        double precision, parameter :: pi = 3.14159265358979d0
+
+        ! The signed toroidal mode remains in m*Omega_b+n*Omega_phi=0 and in the
+        ! Fourier phase.  Only this symmetric search extent is a magnitude.  Using
+        ! signed n here made the native n<0 case produce a negative extent and
+        ! silently discard every otherwise valid resonance contribution.
+        bound = 2.d0*pi*(maxval(abs(dble(m_modes))/abs(dble(n_modes))) &
+            + 1.d0/dble(minval(abs(n_modes))))
+    end function resonant_delphi_bound
+
+    pure logical function canonical_flux_outside_lcfs(psi_star, psi_axis, psi_edge)
+        double precision, intent(in) :: psi_star, psi_axis, psi_edge
+
+        ! Outside means beyond the edge in the axis-to-edge flux direction.  The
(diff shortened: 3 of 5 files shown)


krystophny:
Rebased onto current `main` and kept independent of #104. GitHub Debug and Fast builds pass. The signed-mode change is limited to POTATO resonance bounds/phase handling and its compiled regression; the profile/converter path now lives on main via #102. Follow-up is physical AUG/ITER signed-mode comparison, not more scope in this PR.

krystophny:
**Review verdict: Approve**

Summary: This PR fixes two real bugs in POTATO's resonant torque calculation. First, `delphi_max` (the class root-search bound) was computed with signed toroidal mode numbers, so a native negative `n_tor` produced a negative search extent that silently discarded every resonance. Second, the SOL rejection test `psiast_res < psi_sep` was only valid when flux decreases from axis to edge, wrongly discarding all ITER resonances after `field_eq` selected the opposite GEQDSK gauge. Both are addressed by extracting pure helpers into a new `resonance_mode_bounds_mod` (registered in `potato_base`), and a new ctest `test_signed_toroidal_bound` covers both gauges and both toroidal-mode signs. The gauge-sign logic matches the existing convention already used in `sub_potato.f90` (`(psif-psils)*(psi_sep-psi_axis)>0`). The new bound formula for negative n is correct (`2π(max|m/n|+1/|n|)`).

Findings:
1. [minor] POTATO/SRC/resonance_mode_bounds_mod.f90:13 — `resonant_delphi_bound` divides by `minval(abs(n_modes))`; if `n_tor = 0` (a user-selectable input, `potato_input_mod.f90:34`) this divides by zero. The old code had the identical flaw (`minval(narr)`), so this is not a regression, but the new pure function is the natural place to guard `n == 0`. Impact is limited because a zero toroidal mode is physically degenerate.
2. [minor] No CI configuration exists in the repository (`.github` absent), so test-execution evidence for `test_signed_toroidal_bound` cannot be confirmed here; the test is however correctly registered in `POTATO/CMakeLists.txt` and its assertions are consistent with the corrected semantics.

No other findings; the blank-line addition in `resint_mod` is cosmetic and harmless.

Verdict: Approve — the two fixes are logically correct, the gauge handling matches existing codebase conventions, and the new unit test covers both previously broken cases; the only residual issue (n=0) is pre-existing and non-blocking.

krystophny:
**Review verdict: Approve**

Summary: The PR fixes two signed-mode handling bugs in the POTATO resonant-torque code. (1) `resonant_delphi_bound` (new module) replaces the inline `delphi_max` computation that produced a negative search bound for negative toroidal mode numbers `n`, silently discarding every resonance. The fix correctly takes `abs()` of both `m` and `n`. (2) `canonical_flux_outside_lcfs` replaces the `psiast_res/psi_sep > 1` SOL test with a gauge-invariant "outside the LCFS" comparison that works regardless of flux direction. A new ctest (`test_signed_toroidal_bound`) exercises both functions. The module and test compile cleanly with gfortran (`-Wall -Wextra`), and the test passes.

Findings:
1. [minor] POTATO/SRC/resonant_int.f90:296 and POTATO/SRC/resonance_mode_bounds_mod.f90:24 - The replacement comment states "psif is splined with the axis at zero," yet the new code relies on `psi_axis` being a native (nonzero) gauge value consistent with `psif`/`psiast_res` and `psi_sep`. If `psi_axis` were truly always 0, the new LCFS check is mathematically identical to the old `psiast_res/psi_sep > 1` and fixes nothing; the fix only matters when `psi_axis != 0` (native GEQDSK gauge, including decreasing-flux cases where the old test zeroed all interior points). The comment is internally contradictory and should be reworded to state that `psif`/`psiast_res`, `psi_axis`, and `psi_sep` share the same native gauge. Evidence for that gauge consistency exists (e.g. `potato_resonance_contour.f90:83` computes `psi = psi_axis + rho^2*(psi_sep - psi_axis)` and writes `psi_star` alongside `psi_axis`/`psi_sep`), but the comment undermines confidence in the assumption.

2. [minor] POTATO/SRC/test_signed_toroidal_bound.f90:19-26 - The `canonical_flux_outside_lcfs` tests only use `psi_axis = 0` (both increasing and decreasing flux). With `psi_axis = 0` the new function reduces exactly to the old `psiast_res/psi_sep > 1` behavior, so the test does not actually guard the bug being fixed (nonzero/native axis, e.g. axis=6, edge=5, decreasing flux). Adding a case with nonzero `psi_axis` would make the test meaningful for the reported regression.

3. [minor] POTATO/SRC/resonant_int.f90:60 - A stray blank line was added inside the `resint_mod` module. Cosmetic only; harmless.

Verdict: Approve - the `delphi_max` fix is clearly correct and addresses a real negative-mode bug, and the LCFS check is mathematically sound under the (codebase-supported) same-gauge assumption. The findings are documentation and test-coverage improvements, not blockers.

PR #100, opened 2026-07-26
Add independent direct-field and Fourier contract oracles

krystophny:
## Summary

Adds behavioral probes for native EQDSK coordinates and drift, POTATO Cartesian field/drift/profile/perturbation serializers, the perturbation Hamiltonian, and packed NFP modes.

Production fixes formerly hidden in this branch were extracted to #106 and #107. This PR now contains only probes/oracles plus the required libneo flux-unit pin.

## Validation

- full cumulative `fo` pipeline passes
- probes use analytic or independently integrated physical/coordinate oracles

Stack: #99 -> this PR.

PR #107, opened 2026-07-27
Make POTATO perturbation torque representation-invariant

krystophny:
## Scope

Makes three existing POTATO torque paths representation- and input-consistent:

1. The torque prefactor uses the magnitude of the toroidal mode where the result is a mode-independent magnitude.
2. A positive `enkin_min_over_temp` starts the energy grid above the maximum electrostatic potential; the legacy zero-cutoff grid remains unchanged when the option is zero.
3. The resonance endpoint guard no longer evaluates the declared divergent endpoint before bisection, avoiding an unnecessary separatrix trace.

The signed mode itself is not removed from the resonance phase or mode selection.

## Validation

The diff contains no new standalone test file. It relies on the existing POTATO build, invariant, resonance, and golden-record tests. The three behavior changes should be reviewed as separate follow-up units if this PR is split further.

### DIFF
--- a/POTATO/SRC/resonant_int.f90
+++ b/POTATO/SRC/resonant_int.f90
@@ -319,7 +319,7 @@ subroutine integrate_class_resonances
             !
             ! multiply expression under summation over modes with toroidal mode number, with factor $-\pi^{3/2}/4$
             ! and with reference energy:
-            one_res=one_res*rm3*pi32_over4m*cE_ref
+            one_res=one_res*abs(rm3)*pi32_over4m*cE_ref
             !
             ! A near-tangent root (dresconddx -> 0) gives an infinite
             ! dpsiastdx/dresconddx; for a wall-crossing resonance absHn2=0, so the
@@ -491,6 +491,7 @@ subroutine resonant_torque
         ind_hist,xarr,amat_arr
     use potato_input_mod,  only : nbox, nenerg_input => nenerg, &
         thermen_max_input => thermen_max, &
+        enkin_min_over_temp, &
         adaptive_jperp, npoi_init, nlagr_sampling, &
         eps_sampling, itermax_sampling
     use logging_mod,       only : tee_message
@@ -549,8 +550,17 @@ subroutine resonant_torque
     !
     thermen_max=thermen_max*temp !maximum kinetic energy in units of reference energy
     !
-    ! Energy integration limits:
-    toten_min=phi_elec_min
+    ! Energy integration limits.  Preserve the legacy zero-cutoff grid, including
+    ! its skipped first midpoint.  An explicitly positive minimum instead starts
+    ! above the maximum potential so every sampled orbit has at least the
+    ! requested kinetic energy throughout the field domain.
+    if(enkin_min_over_temp.gt.0.d0) then
+        toten_min=phi_elec_max+enkin_min_over_temp*temp
+        ienerg_begin=1
+    else
+        toten_min=phi_elec_min
+        ienerg_begin=2
+    endif
     toten_max=thermen_max+phi_elec_max
     toten_range=toten_max-toten_min
     !
@@ -584,7 +594,6 @@ subroutine resonant_torque
     torquebox=0.d0
     !
     step_energ=toten_range/dble(nenerg) !integration step over total energy
-    ienerg_begin=2
     !step_energ=0.22521463755624047d0
     !
     omp_threads_env_len=0
--- a/POTATO/SRC/sub_potato.f90
+++ b/POTATO/SRC/sub_potato.f90
@@ -2627,7 +2627,11 @@
   double precision :: xsafe,xdiv,xlo,xhi,xm
   integer :: it
 !
-  if(abs(eval_delphi(xdiv)).le.delphi_max) return
+! xdiv is the boundary-type-declared divergent endpoint.  Do not evaluate it:
+! that defeats this guard by sending find_bounce into an unbounded separatrix
+! trace before the bisection has a chance to move inward.  If the endpoint is
+! actually regular, every interior midpoint stays below the bound and xlo
+! approaches xdiv to the existing interval tolerance.
   if(abs(eval_delphi(xsafe)).ge.delphi_max) return
 !
   xlo=xsafe


krystophny:
Rebased onto current `main`; GitHub Debug and Fast builds pass. This is the minimal production POTATO representation fix extracted from the former oracle bundle #100, with no direct-EQDSK or full-orbit dependency. Follow-up is the independent field-oracle work when that larger feature is resumed.

krystophny:

Summary: This PR ("fix: make POTATO torque representation invariant", 2 files, ~18 lines) wires the previously-ignored `enkin_min_over_temp` namelist parameter into `resonant_torque`'s energy grid (starting the grid above `phi_elec_max` instead of at `phi_elec_min`, and switching the first-midpoint skip accordingly), makes the per-mode torque weight invariant to the sign of the toroidal mode number by using `abs(rm3)` instead of `rm3`, and rewrites the `bound_class_delphi` guard to avoid evaluating the declared-divergent endpoint `xdiv` before bisecting. The `bound_class_delphi` and energy-grid changes are logically sound and well-commented. The main problem is that the wired-up feature contradicts the repo's own committed golden record: `test_resonance_golden.py` expects 3 energy slices generated with `enkin_min_over_temp` "default 0.1", but the code default is 0.0, which produces only 2 slices. I attempted to build POTATO to run the gate but could not (libneo requires Python dev headers, not installable without root in this environment).

Findings:
1. [major] POTATO/SRC/potato_input_mod.f90:40 + POTATO/SRC/resonant_int.f90:556-564 — The new `enkin_min_over_temp` wiring defaults to `0d0`, taking the legacy branch (`ienerg_begin=2`, `toten_min=phi_elec_min`). The committed golden gate `POTATO/test/golden_record_resonance/golden_reslines.json` was explicitly regenerated for "enkin_min_over_temp default 0.1; toten starts at phi_max + 0.1*T" and expects **3** slices for the `nenerg=3` case in `potato.in`. With the actual default 0.0 the code produces only 2 slices (`ienerg=2,3`), so `test_resonance_counts_match_golden` fails on `len(actual) != len(golden)`. The golden `potato.in` does not set `enkin_min_over_temp`, so nothing reconciles this. Note: the test is not registered in POTATO/CMakeLists.txt (its own docstring claims ctest registers `resonance_golden`, but no such `add_test` exists), so it doesn't break automated CI, but the PR ships code that contradicts its own committed golden record, and the intended default (0.1) is clearly what the golden was built for. Fix: either set the default to `0.1`, add `enkin_min_over_temp = 0.1` to `potato.in`, or regenerate the golden for the actual default — and register the test so it actually runs.

2. [minor] POTATO/SRC/resonant_int.f90:322 — `one_res*abs(rm3)` changes torque sign behavior for negative `n_tor` (a no-op for the default `n_tor=1` and all committed test cases since `narr=n_tor`). The "representation invariance" rationale is plausible (modes `(m,n)` and `(-m,-n)` are the same physical harmonic, so a signed `rm3` can spuriously cancel), but this is a physics-behavior change with no test coverage, against the repo's strict TDD mandate. A small test asserting torque invariance under `n_tor -> -n_tor` (with `m -> -m`) would lock in the intent.

3. [minor] POTATO/SRC/sub_potato.f90:2631-2635 — The new `bound_class_delphi` guard is an improvement (it no longer evaluates the declared-divergent `xdiv`, avoiding an unbounded separatrix trace before bisection). One edge case differs from the old behavior: if the *safe* endpoint `xsafe` is already `>= delphi_max`, the routine now returns early leaving `xdiv` untrimmed, so `sample_matrix` may still sample the divergent endpoint and hit `itermax`/drop the class (the very failure mode this routine exists to prevent). For such a fully out-of-range class there are no resonances anyway, so impact is limited, but consider returning the bisected/trimmed value or documenting that an all-out-of-range class is intentionally left untrimmed.

Verdict: Request changes — the wired-up `enkin_min_over_temp` default contradicts the repo's own committed golden gate (2 vs 3 energy slices), and the physics change in `abs(rm3)` ships without the test coverage the repository mandates.

krystophny:

Summary: PR #107 ("fix: make POTATO torque representation invariant") makes three small, coherent changes in `POTATO/SRC/resonant_int.f90` and `POTATO/SRC/sub_potato.f90`. (1) The per-mode weight multiplies by `abs(rm3)` instead of signed `rm3`, so the torque no longer flips sign when `n_tor` is negated — a legitimate invariance fix, since the `m`-sum over a fixed signed `n` would otherwise change sign under the arbitrary Fourier-sign convention of a real perturbation. (2) A new input `enkin_min_over_temp` optionally starts the total-energy integration above `phi_elec_max` so every sampled orbit has positive kinetic energy throughout the domain; the legacy path (`enkin_min_over_temp <= 0`, `ienerg_begin=2`, `toten_min=phi_elec_min`) is preserved exactly. (3) `bound_class_delphi/trim_endpoint` no longer evaluates the boundary-declared divergent endpoint in its guard, avoiding an unbounded separatrix trace before bisection. The logic is sound, but the changes ship with no tests and two minor robustness gaps; build/CI could not be verified in this checkout because `cmake/Util.cmake` (defining `find_or_fetch`) is absent, so CMake configure fails before compilation.

Findings:
1. [major] POTATO/SRC/resonant_int.f90, POTATO/SRC/sub_potato.f90 — No tests were added for any of the three behavioral changes (`abs(rm3)` weighting, the `enkin_min_over_temp` grid path, or the `trim_endpoint` guard). The repository's own CLAUDE.md mandates TDD ("DO NOT WRITE ANY CODE WITHOUT A FAILING TEST FIRST", "TESTS FIRST, ALWAYS"), and the existing `POTATO/SRC/test_*.f90` suite (e.g., `test_sample_matrix_itermax.f90`) covers none of these paths. The `abs(rm3)` and energy-limit changes alter physics results and the trim guard alters root-search behavior, so at least the new cutoff branch and the guard's early-return should be exercised. Required fix: add tests (or a documented justification for waiving the policy) before merge.

2. [minor] POTATO/SRC/resonant_int.f90:568-576 — The new positive-cutoff branch sets `toten_min = phi_elec_max + enkin_min_over_temp*temp` with no check against `toten_max = thermen_max + phi_elec_max`. If `enkin_min_over_temp*temp >= thermen_max`, then `toten_range <= 0` and `step_energ` becomes non-positive, producing a degenerate/empty energy integration (`toten = toten_min + toten_range*xenerg` walks backward or is constant). Required fix: guard `toten_min < toten_max` and handle the empty-range case explicitly.

3. [minor] POTATO/SRC/sub_potato.f90:2659-2662 — The guard condition changed semantics: it now returns early when the *safe* endpoint already satisfies `|eval_delphi(xsafe)| >= delphi_max` (whole interval out of resonance range), whereas the old code evaluated the *divergent* endpoint and would still bisect inward from a within-range `xsafe`. When the entire class interval is out of range, the new code leaves it untrimmed, so `sample_matrix` may grind to `itermax` (class drop, ierr=2) or flood the root search before the no-resonance outcome is recognized downstream. The comment addresses the divergent-endpoint rationale but not this edge; confirm the downstream (`sample_class_doublecount`/`sample_matrix`) tolerates an untrimmed all-out-of-range interval, or trim to `xsafe` in that case.

Verdict: Request changes — the code changes are correct and legacy-preserving, but they ship without the mandatory tests and with two minor robustness gaps (degenerate energy range, untrimmed all-out-of-range class) that should be addressed.

PR #110, opened 2026-07-30
Add coordinate-independent electromagnetic torque kernel

krystophny:
## Scope

Adds a coordinate-independent electromagnetic torque contraction kernel for native complex Fourier amplitudes.

It provides:

- `cross_contraction` for the real-field average of `J1*B2 - J2*B1`;
- `staggered_cross_contraction` for half-mesh `J1/B2` and full-mesh `J2/B1` inputs;
- explicit shape checks that fail closed before invalid mesh contractions.

The kernel preserves the input normalization. It does not perform SI scaling, radial-coordinate conversion, filtering, field reconstruction, or MARS/GPEC file reading. No production reader or adapter is connected by this PR.

## Validation

- Hand-derived complex contraction.
- Explicit full/half-mesh contraction.
- Common complex-phase invariance.
- Invalid half-mesh and full-mesh shape cases are registered as expected-failure tests.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -101,6 +101,7 @@ set(NEO_RT_SOURCES
     src/config.f90
     src/datatypes.f90
     src/driftorbit.f90
+    src/electromagnetic_torque.f90
     src/freq.f90
     src/lib.f90
     src/logging.f90
--- a/src/electromagnetic_torque.f90
+++ b/src/electromagnetic_torque.f90
@@ -1,0 +1,56 @@
+module electromagnetic_torque
+    use, intrinsic :: iso_fortran_env, only: dp => real64
+
+    implicit none
+    private
+
+    public :: cross_contraction
+    public :: staggered_cross_contraction
+
+contains
+
+    pure elemental function cross_contraction(j1, b2, j2, b1) result(value)
+        complex(dp), intent(in) :: j1, b2, j2, b1
+        real(dp) :: value
+
+        value = 0.5_dp*real(conjg(j1)*b2 - conjg(j2)*b1, dp)
+    end function cross_contraction
+
+    function staggered_cross_contraction(j1_half, b2_half, j2_full, b1_full) &
+            result(density)
+        complex(dp), intent(in) :: j1_half(:, :), b2_half(:, :)
+        complex(dp), intent(in) :: j2_full(:, :), b1_full(:, :)
+        real(dp) :: density(size(j1_half, 1))
+        integer :: cell_count, mode
+
+        call require_same_shape(j1_half, b2_half, "half-mesh B2")
+        call require_same_shape(j2_full, b1_full, "full-mesh B1")
+        if (size(j2_full, 2) /= size(j1_half, 2)) then
+            error stop "full/half meshes have different mode counts"
+        end if
+        cell_count = size(j1_half, 1)
+        if (size(j2_full, 1) /= cell_count + 1) then
+            error stop "full mesh must have one more radial point than half mesh"
+        end if
+        density = 0.0_dp
+        do mode = 1, size(j1_half, 2)
+            density = density + 0.5_dp*real( &
+                conjg(j1_half(:, mode))*b2_half(:, mode) - 0.5_dp*( &
+                conjg(j2_full(:cell_count, mode))*b1_full(:cell_count, mode) + &
+                conjg(j2_full(2:, mode))*b1_full(2:, mode)), dp)
+        end do
+    end function staggered_cross_contraction
+
+    subroutine require_same_shape(first, second, label)
+        complex(dp), intent(in) :: first(:, :), second(:, :)
+        character(*), intent(in) :: label
+
+        if (size(first, 1) /= size(second, 1)) then
+            error stop label//" has a different radial size"
+        end if
+        if (size(first, 2) /= size(second, 2)) then
+            error stop label//" has a different mode count"
+        end if
+    end subroutine require_same_shape
+
+end module electromagnetic_torque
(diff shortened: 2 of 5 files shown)


krystophny:
Rebased cleanly onto current `main`; the branch now follows the corrected converter/profile commits without cumulative-stack ancestry. The existing Debug/Fast validation was green before this metadata-only rebase; CI will rerun on the new head. Scope remains the independent electromagnetic contraction kernel and its invalid-input oracle.

krystophny:

## Summary

PR #110 adds a new `electromagnetic_torque` module (`src/electromagnetic_torque.f90`) with two kernel functions (`cross_contraction`, `staggered_cross_contraction`), registers it in the build (`CMakeLists.txt`), and adds three unit tests (one valid, two invalid/`WILL_FAIL` cases). The arithmetic in the tests is hand-verifiable and correct, the Fortran is clean (explicit `use only`, `implicit none`, declarations before executable code, named-tolerance constants), and the tests follow the repo's TDD pattern. The main concern: the module has no production caller anywhere in the tree — it is added to the library build but referenced only by its own tests, and its own documentation defers all real physics (geometry, `B0²/μ0` scaling, coordinate conversion, filtering, edge policies) to an adapter that does not exist in this PR.

## Findings

1. **[major] src/electromagnetic_torque.f90:1 — Dead code with no production caller.** The module is added to `NEO_RT_SOURCES` (CMakeLists.txt:104) and linked into `libneo_rt`, but a repo-wide search shows it is referenced only by its unit tests. Its header explicitly states the results are "not SI torque by themselves" and that geometry/scaling/conversion/filtering/edge policies "belong to an adapter outside this coordinate-independent kernel" — an adapter not present in this PR. As shipped this is unreachable library code whose contract can only be validated self-referentially, not in real use. Required fix: either integrate the kernel with an actual caller (and adapters) in this PR, or gate it behind a feature flag / follow-up issue so the build/link surface doesn't silently grow, and document the intended integration point explicitly.

2. **[minor] test/CMakeLists.txt:77-85 — `WILL_FAIL` tests are fragile pass-throughs.** The two invalid cases correctly trigger `error stop` in `require_same_shape`, but `WILL_FAIL` turns any exit-status failure into "pass", including a future crash or a silent no-op regression. Acceptable for validating fail-closed guards, but consider asserting on the specific error (e.g., capturing stderr) so the tests fail loudly if the guard is ever removed or mis-reached. Not blocking.

3. **[minor] src/electromagnetic_torque.f90:27-52 — Contract depends on undocumented mesh conventions.** The 1/2 real-field factor, the extra 1/2 for endpoint averaging of full-mesh products, and the requirement that the full mesh has exactly one more radial point than the half mesh are documented only in prose. With no caller to anchor the convention, a future adapter could silently mismatch the sign/averaging convention with no test coverage. Recommend naming the 1/2 factors as named constants with comments (per the repo's magic-number rule) and pinning conventions with a second independent fixture. Non-blocking.

## Verdict

**Request changes** — the kernel and tests are correct and well-written, but shipping a library module with no production caller and an adapter that does not exist makes this dead code; it should be integrated or explicitly staged before merge.

krystophny:
**Review verdict: Approve**

# Review: PR #110 — Electromagnetic-torque contraction kernel

**Summary:**
This PR adds a small, self-contained Fortran module (`src/electromagnetic_torque.f90`) implementing an algebraic electromagnetic-torque contraction kernel for complex Fourier harmonics, plus two unit-test programs and CTest registrations. The module is added to `NEO_RT_SOURCES` (thus built into the `neo_rt` library), and the tests are auto-globbed by the existing `file(GLOB TEST_SOURCES "test_*.f90")` pattern in `test/CMakeLists.txt`, so both `test_electromagnetic_torque.x` and `test_electromagnetic_torque_invalid.x` are compiled and linked automatically. The two `invalid_*` tests are correctly marked `WILL_FAIL TRUE`. The PR is purely additive with no production callers yet; the module's contract is documented in its header comment.

**Findings:**

1. [minor] `src/electromagnetic_torque.f90` — The new kernel has no production caller; it is dead code until the documented "adapter" (geometry scaling, B0²/μ0, radial-coordinate conversion, filtering, edge policies) is implemented. This is acceptable as an incremental kernel-first contribution and is explicitly acknowledged in the header comment, but the PR cannot affect physics output until that adapter lands. No fix required for merge; just tracking that this is a standalone building block.

**Checks performed:**
- Module compiles cleanly with `gfortran`.
- Hand-derived expected value `7.875_dp` in `test_hand_derived_mode_contraction` independently verified: `sum` over both modes of `0.5*Re(conjg(j1)*b2 − conjg(j2)*b1)` = 7.875.
- `test_explicit_full_half_stagger` expected `0.5_dp` independently verified from the explicit arrays, including the adjacent-endpoint averaging (`j2_full(:n)` and `j2_full(2:)`).
- `test_common_complex_phase` correctly checks invariance under a common complex phase (unit-magnitude `0.6+0.8i`), validating the Fourier-sign convention independence claim.
- Shape-validation logic in `staggered_cross_contraction` is sound: the `half` invalid case trips the half-mesh B2 shape check, and the `full` invalid case trips the full-mesh B1 shape check, both via `error stop` (non-zero exit), consistent with the `WILL_FAIL` test properties.
- No wildcard imports, `implicit none`, `use only`, all declarations before executable statements, and named constants used for tolerances — consistent with the repo's mandatory coding standards.
- Existing `test/CMakeLists.txt` glob will pick up the new test sources; no registration was missed.

**Verdict:** Approve — the diff is small, well-scoped, compiles cleanly, and every test expectation was independently verified against the implementation; it introduces no risk to existing behavior.

PR #111, opened 2026-07-30
Add strict GPEC torque boundary

krystophny:
## Scope

This PR is stacked on #110 and contains only the incremental GPEC boundary contract.

It adds `gpec_torque_contract`, which reads exactly one finite toroidal torque from a native `GPEC_CONTROL` file when the toroidal mode is supplied by the caller. It rejects zero mode, malformed or missing torque records, and requests for a volume perturbed-current vector.

The README documents the limitation explicitly: GPEC’s scalar control output is an independently integrated torque, not the vector/current data required by the coordinate-independent electromagnetic `JxB` kernel. No GPEC-to-MARS or full-volume reconstruction is claimed.

## Validation

- Valid `GPEC_CONTROL` fixture with `n=3`.
- Missing-torque fixture.
- Zero-mode rejection.
- Explicit rejection of unavailable volume-current data.

Merge order: #110 first, then this stacked contract.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -103,6 +103,7 @@ set(NEO_RT_SOURCES
     src/driftorbit.f90
     src/electromagnetic_torque.f90
     src/freq.f90
+    src/gpec_torque_contract.f90
     src/lib.f90
     src/logging.f90
     src/magfie.f90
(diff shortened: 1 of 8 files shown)


krystophny:
Rebased after #110 onto the new current-main kernel head. This remains a one-commit child PR for the strict GPEC boundary contract; no direct-EQDSK or full-orbit code is included. CI will rerun against the rebased stack.

krystophny:
Manual workflow-dispatch validation completed successfully at rebased head 6e5c357: build (Fast) and build (Debug) both passed. This stacked PR targets #110, so the normal workflow trigger (pull requests targeting main) does not attach a check suite to it.

PR #116, opened 2026-08-07
Separate perturbation input format from equilibrium format

krystophny:
## Summary

- add `inp_swi_pert` so the standalone equilibrium and perturbation files can select their formats independently
- retain legacy behavior by inheriting `inp_swi` when `inp_swi_pert` is negative
- reject unsupported perturbation `.bc` formats before the reader can use an undefined column layout
- extend the manufactured Fourier-field oracle to exercise both a chartmap-equilibrium/Strumberger-perturbation configuration and the inherited legacy configuration

This is an input-routing change only. It does not change orbit equations, radial normalization, Fourier signs, torque signs, or POTATO physics.

## Validation

Validated from exact commit `11ba86e5fa475003125f9186baa4d5470455b1c7` in a clean checkout on `faepkub4`:

- `git diff --check`
- `fo build`
- manufactured mixed-reader Fourier oracle: exact agreement with the analytic field
- `fo test --all`: 12/12 passed (Release)
- `fo build --debug` and `fo test --all`: 12/12 passed
- `make pytest`: 15/15 passed, including all golden-record cases

## Existing repository-tooling baseline

The aggregate `fo` command aborts before building because its static parser crashes on existing legacy fixed-form `write` statements. Running its build/test gates separately succeeds. `fo lint` completes and reports the existing two unused imports and 76 legacy warnings; none points to a line added by this PR.


krystophny:
Rebased onto current `main`; the mixed equilibrium/perturbation input-format routing remains independent of orbit, torque, sign, and generated-kernel changes. Existing Release/Debug, focused Fourier-oracle, and pytest validation remains the relevant gate.

PR #117, opened 2026-08-07
Initialize minor radius from Boozer chartmaps

krystophny:
## Scope

Initializes the standalone Boozer chartmap minor radius from the chartmap-provided `aminor` value instead of leaving the standalone geometry without that scale.

The chartmap input test is updated to cover the supplied minor-radius value and the existing default/compatibility path. No radial-coordinate conversion or torque convention is changed here.

## Validation

- Updated `test_chartmap_input` checks the chartmap minor-radius field.
- The PR is limited to standalone chartmap initialization and its input oracle.


krystophny:
Rebased onto current `main`; this is an input-state initialization fix with an independent chartmap-vs-`.bc` radius oracle. It does not overlap the converter, resonance, torque, or full-orbit PRs.

PR #118, opened 2026-08-07
Stabilize direct full-FOW generated kernels and runtime contracts

krystophny:
## Summary

- enforce type-correct, case-aware Fortsym kernel interfaces and regenerate the checked-in cylindrical, full-FOW, and POTATO kernels
- derive the finite-aspect circular-EQDSK volume metric with Fortsym and strengthen the radial metric oracle on resolved surfaces
- repair certified interval-root subdivision and independent behavioral fixtures needed by the direct full-orbit path
- add a certified class-topology provider seam and make the production EQDSK factory fail closed when no physical topology provider is available

## Physics scope

This PR does not claim that POTATO physics was wrong. The Fortsym certificates prove that the Buchholz and POTATO action normalizations map to the same physical magnetic moment. The corrected tests use

    mu_phys = J_K * abs(q) / (m * c)

for the Buchholz action convention.

The production topology atlas is deliberately not invented here. The runtime reports status 1004 when certified physical topology is unavailable, and the regression test verifies that behavior together with the successfully initialized field, profile, perturbation, and wall evidence.

## Verification

Run on faepkub4 from a pushed clean Git commit:

- Debug CTest: 65/65 passed with OMP_NUM_THREADS=1
- Release CTest: 65/65 passed with OMP_NUM_THREADS=1
- Fortsym direct full-FOW suite: 107/107 proved; every generated file matched byte-for-byte
- Fortsym finite-aspect circular-torus metric suite: 4/4 proved; generated kernel matched byte-for-byte

## Follow-up boundaries

The exact Fortsym Eq. 13 cut jet and the certified production topology-atlas provider remain separate changes. This PR provides the green, fail-closed foundation they require.


krystophny:
Closing this generated full-FOW/direct-transport stack as deferred. It is a large feature/verification bundle, not an independently mergeable small fix, and its generated-kernel work will be refiled only with the complete direct-field/full-orbit design.

PR #121, opened 2026-08-10
Report orbit failures from get_matrix_doublecount (diagnosis + reproducer)

phizenz:
## Summary

`get_matrix_doublecount` could not tell its caller that an orbit evaluation had failed. This PR makes the failure observable and converts the two callers that silently depended on the old convention. 

## The defect

`get_matrix_doublecount` takes no arguments and only writes `amat` at the very end. Both error returns — starter failure, and `find_bounce` hitting `ierrfield` when the orbit leaves the field domain or crosses the wall — return before that point, leaving `amat` holding **the previous node's values**.

`sample_matrix` calls it with no status argument, so a failed node is indistinguishable from a good one. The stale values read as structure, and the refinement bisects toward them.

## What this PR changes

- `ierr_get_matrix` added to `sample_matrix_mod` (threadprivate, matching the surrounding class-grid state).
- `get_matrix_doublecount` sets it on entry and, on every failure path, sets it to 1 and clears `amat`.
- `bound_class_wall/orbit_lost` and `bound_class_delphi/eval_delphi` now test the flag instead of the stale-value sentinel.

That last point is the part most worth reviewing. **Two** routines depended on the old convention, not one. Both poisoned `amat(3,1)` with `huge()` and tested whether it survived the call: `orbit_lost` to detect wall losses, `eval_delphi` to make a failed orbit read as beyond `delphi_max`. Clearing `amat` without updating them would have silently disabled wall trimming **and** X-point trimming, with no diagnostic — the bisections would simply have concluded that nothing was ever lost.

### DIFF
--- a/POTATO/CMakeLists.txt
+++ b/POTATO/CMakeLists.txt
@@ -86,6 +86,19 @@ add_test(NAME test_find_all_roots_bracketed
    COMMAND test_find_all_roots_bracketed.x
 )
 
+add_executable(test_failed_orbit_nodes.x
+   SRC/test_failed_orbit_nodes.f90
+)
+# sample_matrix pulls sub_potato (module state), which calls into a
+# profile_input implementation.  potato_base is intentionally incomplete there,
+# so the two archives must be rescanned as a group rather than linked in order.
+target_link_libraries(test_failed_orbit_nodes.x
+   "$<LINK_GROUP:RESCAN,potato,potato_base>"
+)
+add_test(NAME test_failed_orbit_nodes
+   COMMAND test_failed_orbit_nodes.x
+)
+
 add_executable(test_wall_loss.x
    SRC/test_wall_loss.f90
 )
--- a/POTATO/SRC/CMakeLists.txt
+++ b/POTATO/SRC/CMakeLists.txt
@@ -1,5 +1,5 @@
 add_library(potato_base
-  ${LIBNEO_SOURCE_DIR}/src/libneo_kinds.f90
+  ${LIBNEO_SOURCE_DIR}/src/util/libneo_kinds.f90
   logging_mod.f90
   wall_loss_mod.f90
   potato_input_mod.f90
--- a/POTATO/SRC/sample_matrix.f90
+++ b/POTATO/SRC/sample_matrix.f90
@@ -114,8 +114,13 @@
   DO
     iter=iter+1
     IF(iter.GT.itermax) THEN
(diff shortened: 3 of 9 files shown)


krystophny:
# Split plan for this bundle

I am splitting #121 by behavior on current `main` (`6b399ca`), because the present branch mixes independent fixes and one unsafe policy change:

1. **Failed-orbit handling**: `dba495e` exposes an explicit `ierr_get_matrix` status and updates wall/X-point trimming to consume it. The accompanying `7b37eaf` reproducer is intentionally red until adaptive `sample_matrix` refinement skips invalid orbit nodes. I am carrying those as one complete, tested PR so clearing stale `amat` cannot regress the callers. This is related to the fail-closed accounting discussion in #80, but is a smaller POTATO-only fix.
2. **NEO-2 profile import gauge**: `62f8179` plus `18a7d4b` will become a separate utility PR. It fixes the toroidal-vs-poloidal flux normalization in `import_neo2_profiles.py`. The existing #102 work covers the text NEO-RT-to-POTATO converter and explicit `s_tor -> s_pol` mapping; the HDF5 importer is a distinct input path, while the duplicate `python/neort_to_potato.py` in this branch will not be copied into #102 without a separate API/test decision.
3. **Not promoted**: `35c5313` makes an unresolved adaptive quadrature silently succeed at `itermax`; that weakens the error contract and is not a fix. `55fb0cc` is already present in the current POTATO CMake path. `test_bounce_integrand.f90` is an unregistered diagnostic probe and belongs with the frequency/resonance diagnostics in #90 rather than a production fix.

The replacement PRs will state their test evidence and follow-up boundaries explicitly. This mixed PR will be closed as superseded once those replacements are published.

krystophny:
Replacement drafts are now published:

- #122 — NEO-2 HDF5 profile import with the corrected EQDSK poloidal flux gauge.
- #123 — complete failed-orbit handling: explicit status plus adaptive-grid validity tracking.

Both are current-main based and document their tests and follow-up boundaries. The remaining #121 commits are intentionally not promoted: accepting `itermax` would weaken the error contract, the libneo path is already present on current main, the bounce probe is unregistered diagnostic code, and `python/neort_to_potato.py` overlaps the converter API in #102.

krystophny:
Superseded by the current-main atomic PRs #122 and #123. The remaining mixed commits were intentionally not promoted; see the split analysis comments above.

PR #122, opened 2026-08-10
POTATO: use the EQDSK poloidal flux gauge for NEO-2 profile import

krystophny:
## Meaning

This is the NEO-2 HDF5 importer portion of #121.  It generates `profile_poly.in` from NEO-2 profiles and derives the electric field with POTATO’s EQDSK poloidal flux span, `(PsiedgeVs - PsiaxisVs)`, rather than NEO-2’s toroidal `psi_pr_hat*Bref`.  It also maps the input radial coordinate from normalized toroidal flux to normalized poloidal flux before fitting.

Using the toroidal gauge inflated `Omega_E` by about 1.945 in the AUG 30835 case and displaced the radial profile.  The correction is an input-contract fix; it does not alter NEO-RT or POTATO orbit dynamics.

## Why this is separate from #121

The unregistered `test_bounce_integrand.f90` probe was not copied here.  Frequency/resonance diagnostics belong with #90 and the current benchmark work.  The separate `python/neort_to_potato.py` file in #121 is also not copied: #102 already carries the validated text NEO-RT-to-POTATO converter and explicit `s_tor -> s_pol` mapping.  Combining two converter APIs would create duplicate paths before their fixture contract is agreed.

## Evidence

- Standalone syntax/import checks for the utility.
- `POTATO/test/profile_import/test_import_neo2_profiles.py` uses a synthetic HDF5 input and fake EQDSK/libneo mapping to verify the `s_tor -> s_pol` call and the EQDSK poloidal-gauge potential.
- The behavioral test passes.

## Follow-up boundary

Before merge, validate this HDF5 path on AUG 30835 and compare the resulting canonical frequencies/resonance lines against NEO-RT, POTATO, and SIMPLE.  If the project wants one public converter interface, that follow-up should reconcile this HDF5 path with #102 rather than add a third wrapper.  No torque sign is changed here.

### DIFF
--- a/POTATO/python/import_neo2_profiles.py
+++ b/POTATO/python/import_neo2_profiles.py
@@ -1,0 +1,198 @@
+#!/usr/bin/env python3
+"""
+Generate POTATO profile_poly.in from NEO-2 HDF5 output.
+
+Reads neo2_out.h5 and computes the electrostatic potential phi_e from the
+ExB rotation (MtOvR):
+
+    Er(s) = sign_theta * psi_pr * vth(s) * MtOvR_ion(s) / C_CGS
+
+where psi_pr must be the POLOIDAL flux difference in POTATO's own gauge,
+psi_pr = (PsiedgeVs - PsiaxisVs) * 1e8 Mx from the EQDSK header, because
+POTATO's phielec_of_psi divides dPhi/ds_pol by exactly that quantity.
+Using NEO-2's psi_pr_hat*Bref here instead (that is the TOROIDAL flux
+derivative, -3.585e7 Mx) inflates Omega_ExB by x1.945 at every radius --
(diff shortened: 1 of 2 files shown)


krystophny:
Split from #121: this PR contains only the NEO-2 HDF5 profile-input contract. It is complementary to the text converter in #102, not a replacement for it. The remaining follow-up is real AUG 30835 validation and deciding whether both input paths should share one converter interface; no torque sign is changed here.

krystophny:
The squash content is on `main` as `e5cb9c1`, with Git author Philipp Zenz <philipp.zenz@tugraz.at>. GitHub’s merge endpoint recorded the first squash as Christopher despite the author request, so the identical tree/message was safely replaced with a force-with-lease; parent remains `6b399ca`.

PR #123, opened 2026-08-10
POTATO: handle failed orbit nodes without stale interpolation data

krystophny:
## Scope

Prevents failed POTATO orbit evaluations from being treated as valid interpolation data during adaptive sample-matrix refinement.

The production path records the status of each `get_matrix` call, marks failed orbit nodes invalid, skips interpolation windows containing invalid nodes, and keeps the last completed grid when the refinement iteration cap is reached. This avoids stale `amat` values driving further refinement toward an orbit that left the valid field domain.

The existing sample-matrix nonconvergence status from merged main is preserved; this PR adds the per-node failure handling needed by the refinement consumer.

## Validation

Adds and registers `test_failed_orbit_nodes`, an independent stub oracle with a deliberately failed interval. It verifies that refinement does not chase the failed interval or insert nodes into it.

### DIFF
--- a/POTATO/CMakeLists.txt
+++ b/POTATO/CMakeLists.txt
@@ -88,6 +88,19 @@ add_test(NAME test_find_all_roots_bracketed
    COMMAND test_find_all_roots_bracketed.x
 )
 
+add_executable(test_failed_orbit_nodes.x
+   SRC/test_failed_orbit_nodes.f90
+)
+# sample_matrix pulls sub_potato (module state), which calls into a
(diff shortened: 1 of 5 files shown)


krystophny:
Split from #121: the status plumbing and adaptive-grid validity tracking stay together because the original #121 reproducer fails if `amat` is merely cleared. The unresolved-grid `itermax` behavior was intentionally excluded. Follow-up is a production AUG 30835 orbit-class/resonance scan and the broader accounting scope in #80; #90 remains the independent resonance-ledger diagnostic.

PR #125, opened 2026-08-10
POTATO: seed the Poincare cut at the equilibrium midplane

krystophny:
## Summary

Seed the POTATO Poincare-cut search at the actual EQDSK grid midplane, `0.5*(zet(1)+zet(nzet))`, instead of assuming `Z=0`. The seed is used for the initial, forward, and backward cut searches.

This matters for vertically shifted equilibria, including non-circular cases such as AUG 30835 and ITER equilibria whose supplied grid is not centered at zero.

## Validation

- Clean current-`main` POTATO configure/rebuild completed successfully.
- `potato.x` builds with the change.
- The change is limited to the Poincare-cut seed; it does not alter orbit equations, signs, normalization, or torque formulas.

## Scope

Small input/initialization fix. A production AUG 30835 validation run remains follow-up work.

### DIFF
--- a/POTATO/CMakeLists.txt
+++ b/POTATO/CMakeLists.txt
@@ -88,6 +88,19 @@ add_test(NAME test_find_all_roots_bracketed
    COMMAND test_find_all_roots_bracketed.x
 )
 
+add_executable(test_poicut_midplane.x
+   SRC/test_poicut_midplane.f90
+)
+target_link_libraries(test_poicut_midplane.x
+   potato_base
+   potato
+)
+add_test(NAME test_poicut_midplane
+   COMMAND test_poicut_midplane.x
+   ${CMAKE_CURRENT_BINARY_DIR}/shifted_convexwall.dat
+   WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}/test/golden_record_resonance
+)
+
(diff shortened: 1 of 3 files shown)


krystophny:
Filed as an independent POTATO initialization fix. Debug CI passes; the Fast build is still running. The scope is the EQDSK grid midplane seed for vertically shifted equilibria; AUG 30835/ITER production validation is follow-up work.

krystophny:
Added an independent behavioral regression test in 413c4b7: `test_poicut_midplane` translates the equilibrium grid and checks both Poincare-cut endpoints follow the translated geometric midplane.

Validation: fixed branch passes the CTest; the same test applied to `origin/main` fails, so this is not merely a source/state assertion.

krystophny:
CI is green on the rebased head 413c4b7: Debug and Fast both pass, in addition to the local shifted-grid CTest and the baseline-negative check.

krystophny:
Closing as superseded by merged #127. #125 targeted the previous midplane-seeded tracing algorithm; #127 replaced it with magnetic-axis tracing, so rebasing this patch would reintroduce duplicate/incompatible production logic. The shifted-equilibrium coverage should be retained only if a test is written against the current axis-tracing implementation.

PR #126, opened 2026-08-10
POTATO: keep custom root grids inside the search bracket

krystophny:
## Scope

Keeps POTATO custom root grids inside the caller-provided search interval.

The bracketed root finder now clips its generated grid to the requested lower and upper bounds instead of allowing the custom grid to extend beyond them. The existing root-search behavior is retained for ordinary grids.

## Validation

Updates `test_find_all_roots_bracketed` with bounded custom-grid cases and verifies that returned roots stay within the requested interval.

### DIFF
--- a/POTATO/SRC/find_all_roots.f90
+++ b/POTATO/SRC/find_all_roots.f90
@@ -32,7 +32,7 @@
 !
   implicit none
 !
-  integer :: i,iter,nsearch,ierr,ndummy,kx,kxc,k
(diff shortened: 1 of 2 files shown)


krystophny:
Filed as an independent POTATO root-search fix. Debug CI passes; the Fast build is still running. The regression records evaluations outside `[x1in,x2in]` and covers both root-search entry points; no orbit or torque formulas change.

PR #127, opened 2026-08-10
POTATO: run on ITER equilibria

phizenz:
## Summary

POTATO could not produce a torque on ITER CHEASE equilibria. Three independent assumptions baked in from AUG validation had to be lifted: an up-down symmetric magnetic axis, a device-sized (R,Z) grid, and a negative `psi_sep`. Each failed differently — non-convergence, a stalled fixpoint search, and a silently zeroed torque.

### Poincare cut assumed the axis is on the midplane (`aceaba3`)

`find_poicut` seeded its Newton search at `R=rtf, Z=0`. That lands on the right branch of the `grad(psi) x grad(B) = 0` curve only when the magnetic axis sits on the midplane — true for AUG, not for ITER where it is tens of centimetres off. The iterations then never converge.

The axis is now located first and independently, by Newton iteration on `grad(psi) = 0` seeded from the extremum node of the splined flux, and the cut is traced outward from there in both directions (`trace_poicut_end`, `poicut_z_of_R`).

Two hardening changes ride along, both previously silent failures: iteration loops that could exhaust their count now `error stop` instead of returning a half-converged point, and the `rho_pol` boundary search is clamped to the traced cut so the Lagrange interpolation in `get_poicut` stays bounded.

### g-file grid resolution did not scale with device size (`79ad342`)

`boozer_to_efit` emitted a fixed 200x200 (R,Z) grid. What the psi splines and the separatrix fixpoint search actually see is the *spacing*, not the point count: 200x200 gives `hz = 4.4 cm` on an ITER-sized box against the ~0.6 cm of the AUG files POTATO was validated on, and the fixpoint Newton then stalls near 1e-4 against a 1e-12 relative tolerance.

`nr`/`nz` are now derived from a 0.75 cm target spacing, with 200 kept as a floor so small boxes are unaffected.

### SOL resonance test was hardcoded to one sign convention (`3a8d2f5`)

The check that zeroes resonances outside the separatrix compared `psiast_res` against `psi_sep` directly. `psif` is splined with the axis at zero, so `psi_sep` is negative for AUG but positive for the ITER CHEASE files, and the bare inequality is inverted for one of them. On ITER it zeroed every resonance, and with it the entire torque.

Now compared as normalized flux, which holds under either convention.

## Please review separately

`804c4a3` (`nsearch_min` 100 -> 400) is **not ITER-specific**. It widens the `find_all_roots` bracketing scan that decides how many orbit classes exist and where their boundaries lie. At 100 equidistant brackets over the whole cut, a class narrower than ~1% of it can fall between two samples and never be found — near-axis potato classes are the narrow ones. But it changes class discovery on every equilibrium, AUG included, so it may shift existing results.


krystophny:
Audit status: this branch is based on the pre-current-main history and is not mergeable as one PR. Comparing its tip with current `main` shows 46 changed paths, including deletion of the merged POTATO converters/tests and unrelated MARS-JXB/GPEC artifacts; it would also reintroduce the superseded `python/neort_to_potato.py` stack.

I converted this PR to draft while preserving the branch. The ITER work needs to be split after rebasing: (1) Poincare-cut/axis support is a substantial follow-up, (2) Boozer-to-EFIT grid sizing is a separate small fix, (3) the psi-separatrix sign fix is a separate small fix, and (4) the 400-bracket root-scan change needs independent AUG/ITER evidence before filing. The old profile/converter/probe and itermax commits are already classified as duplicate, diagnostic, or unsafe in the cleanup audit.

PR #129, opened 2026-08-11
Replace thin-orbit separatrix frequency extrapolation

krystophny:
## Scope

This PR is the current integration vehicle for the validated thin-layer orbit treatment. It replaces the old two-point logarithmic extrapolation and fixed boundary margins with a generated, geometry-aware separatrix model and machine-adjacent resonance search.

- fit the bounce-time integral and unnormalized drift numerator in the explicit distance `abs(eta-etatp)/etatp`, then evaluate the finite drift ratio `N/tau`
- generate the asymptotic kernel through the NEO-RT-owned fortsym generator
- choose an admissible fit window from resolved positive/finite data, with a principled leading-log fallback for under-resolved correction terms
- search resonance brackets logarithmically in distance to the trapped-passing boundary, through the adjacent representable pitch value without a physical `epst`/`epsp` cutoff
- retain the current safeguarded root solver, independent oracle, diagnostic sources, and unequal-periodic-well geometry coverage from `main`
- preserve the separate passing magnetic-drift gate and the current Fast-build configuration

The finite precession limit follows from fitting the unnormalized bounce integrals before taking their ratio; this is the behavior checked against the direct barrier oracle and the literature-derived elliptic-integral limit.

## Provenance / validation

The final frequency and resonance comparison plots were generated from this thin-layer branch during the prior Codex session; the branch history records machine-adjacent curves and the corrected `N/tau` drift treatment. The previously stacked resonance PR is already merged into this branch; current `main` is integrated at merge commit `b91f292`.

- Fast CTest: 16/16 passed
- Debug CTest: 16/16 passed
- Python/golden suite: 15/15 passed
- generated kernel regeneration and independent separatrix/orbit-class/resonance oracles pass
- no physical cutoff was reintroduced; exact singular endpoint evaluation remains excluded only by representability

Primary references used for the endpoint behavior: DLMF 19.12; Connor–Hastie–Martin; Roach–Connor–Janjua; Mackenbach et al.; Rodríguez–Mackenbach. The smooth nondegenerate-barrier assumption and multi-well limitations remain documented explicitly.


krystophny:
**Review verdict: Approve**

Summary: This PR replaces NEO-RT's ad-hoc near-separatrix log extrapolation with a fitted asymptotic "thin-orbit" model (new `src/separatrix.f90` plus a committed fortsym-generated kernel `src/generated/neort_separatrix_frequency_symbolic.f90`), and adds per-well magnetic-well inventory (`src/orbit_classes.f90`) so multi-well Boozer surfaces get per-class trapped splines instead of one global Bmin/Bmax pair. The resonance search is re-parametrised in log-distance from the trapped–passing boundary, `driftorbit_root` is refactored onto a reusable safeguarded Newton/bisection `solve_bracketed_root`, and the `magdrift`/`magdrift_passing` switches are made consistent for passing orbits. The `epst`/`epsp` exclusion layers are dropped in favour of `ieee_next_after` boundary values, the golden record is regenerated with documented rationale, and an optional fortsym generator is added. The change is large but coherent and unusually well tested.

Findings:
1. [minor] `test/test_resonance_bracket.f90` — the independent plain-bisection oracle that cross-checked the production physical root against a separate solver was removed, replaced by a synthetic analytic-root oracle for `solve_bracketed_root` plus a residual-convergence check on the real root. Coverage remains meaningful, but the physical resonance root is no longer verified against an independent root finder over the same bracket; consider re-adding a bisection comparison for the physical root.
2. [minor] `src/orbit_classes.f90` (threadprivate `classes`/`global_*`) — the new per-well state is `threadprivate` but has no per-thread initialisation guard, unlike the pre-existing frequency state in `src/freq.f90` which lazily re-inits via `freq_thread_init_state`. In this checkout there is no OpenMP parallel region, so it is inert; but if a worker thread ever calls `select_orbit_class`/`orbit_class_count` without first running `init_orbit_classes` on that thread, it reads uninitialised copies. A `!$omp single` guard or a per-thread init flag would make this robust for the libneo-external parallel path.
3. [minor] `test/golden_record/REGENERATION.md` — the regenerated `golden.h5` (blob changed, same byte size) documents a physics/reference correction for case `0p300` and a scale-relative `rtol=5e-3` comparison bar. This is transparently disclosed and argued, but reviewers should note the golden comparison is now looser than the historical `rtol=1e-8/atol=1e-15` bar. No CI workflow exists in this checkout (`.github` absent); the 15/15 ctest / ripple-plateau / 9/9 golden evidence is the author's local run recorded in the doc.

No blockers or major findings. The model is gated correctly (`x <= xscale` covers up to and beyond the innermost spline knot, so there is no unsplined gap), the removed `epst`/`epsp` constants have no remaining references, `roots(:,3)` (no longer filled) is unread, and the `magdrift → magdrift_passing` change in `Om_ph` is consistent with `Om_tB` and the earlier `magdrift_passing` feature.

Verdict: Approve — the physics change is deliberate, internally consistent, and backed by strong tests (manufactured oracle, direct `do_magfie` barrier-top oracle, window-invariance and direct-orbit checks, per-well inventory oracle) plus a reproducible committed generated kernel. The remaining items are non-blocking suggestions.

PR #131, opened 2026-08-12
Optimize safeguarded resonance root search

krystophny:
## Summary

- Refine each coarse resonance bracket with safeguarded Newton/secant steps and forced midpoint contraction instead of rescanning all 100 eta cells per root.
- Keep residual and eta-width convergence criteria separate so near trapped--passing roots are not accepted from a poorly conditioned residual alone.
- Reuse the passing-branch `Omega_theta` evaluation and retain finite-value, endpoint-Jacobian, exact-grid, and sentinel guards.
- Add a bisection root-selection oracle at two surfaces, a centered Jacobian check away from the non-differentiable separatrix, and derivation/literature documentation.

The bounce search is intentionally unchanged: Fortnum VODE already locates event roots with Illinois safeguarded interpolation on its dense output. Changing the trapped acceptance or passing `+2*pi` contract changed the golden physics and is not part of this PR.

## Verification

- `fo build` — passed.
- `fo test` — 13/13 passed.
- `pytest test/ripple_plateau/test_ripple_plateau.py test/golden_record/test_golden_record.py` — 11/11 passed.
- `test_resonance_oracle` checks both `s=0.5` and `s=0.039063`, both signs of parallel velocity, passing/trapped regions, 16 velocities, and `mth=-12..12` against plain bisection.
- `test_resonance_bracket` passes with no stderr diagnostics.
- A representative `test_misc driftorbit` workload was about 0.20 s on this feature build versus 0.22 s on post-merge main; the high-level workload is bounce-dominated, so the root-path gain should be measured again on the production cluster matrix.

`fo lint` and the repository-wide format check still report pre-existing unrelated POTATO/NEO-RT warnings and unused imports; no new lint error was introduced by this patch.

## References

- Brent root-finding report: https://maths-people.anu.edu.au/~brent/pub/pub006.html
- Alefeld, Potra & Shi, Algorithm 748: https://dl.acm.org/doi/10.1145/210089.210111
- Boost TOMS 748 documentation: https://www.boost.org/doc/libs/latest/libs/math/doc/html/math_toolkit/roots_noderiv/TOMS748.html
- Albert et al. NEO-RT resonance condition: https://arxiv.org/abs/1607.04665


krystophny (commented):
**Review verdict: Comment**

## Review of PR #131: "Optimize safeguarded resonance root search"

**Summary:** This PR replaces the bisection-based refinement in `driftorbit_root` with a safeguarded Newton/secant/midpoint hybrid that keeps the enclosing-bracket invariant, and refactors residual/Jacobian evaluation into a shared `resonance_value` that inlines the passing-branch `Om_ph = Om_tE + Om_th/iota (+ Om_tB)` identity to avoid a redundant `Om_th` call. It adds exact-grid-point root handling in `driftorbit_coarse`, a new `test_resonance_oracle` cross-checking production roots and Jacobians against an independent bisection oracle, and extends `test_resonance_bracket`. The passing-branch expansion exactly matches `Om_ph`'s passing path (the `supban` case applies only for `eta > etatp`, which still routes through `Om_ph`), and the returned Jacobian formula is unchanged, so production roots converge to the same bracket-invariant root as before. The change is well-conceived and the oracle test is strong; only minor issues remain.

**Findings:**

1. **[minor] `src/resonance.f90:103`** — `use logger, only: debug, warning` imports `debug`, which is never referenced anywhere in the module. Dead import; emits a compiler warning and could break a `-Werror` build. Remove `debug` from the import list.

2. **[minor] `src/resonance.f90:13-14, 34`** — The comment "each residual evaluation computes the expensive poloidal frequency only once" (also in `doc/differentiation.md`) holds only for the passing branch. In the trapped branch (`eta > etatp`), `resonance_value` calls `Om_th` unconditionally at the top and then calls `Om_ph`, which calls `Om_th` again — two evaluations. Code is correct, but the stated optimization rationale is overstated.

3. **[minor] `src/resonance.f90` eta_tol exit branch** — When the bracket width reaches `eta_tol ≈ 4·ε·|eta|`, the midpoint is accepted based only on `ieee_is_finite(res)` and `dresdeta /= 0`, ignoring the residual magnitude. Safe for continuous sign-changing residuals (root pinned to machine precision), but it partially defeats the PR's stated residual+eta-width pairing for avoiding a "visibly displaced root" in a flat/discontinuous residual. Low risk given oracle coverage, but worth a guard or comment.

4. **[minor] No output-level regression guard** — `driftorbit_coarse` now always counts exact grid-point zeros (previously some were lost when the left neighbor had the same sign), and the returned `dresdeta` feeds `transport.f90`'s `abs(eta_res(2))` denominator and the hard-root Jacobian. Oracle tests verify root/Jacobian agreement with bisection, but there is no golden-record check that torque outputs are unchanged. The repo also contains no CI workflow, so the new tests' pass status cannot be confirmed from repository CI evidence.

**Verdict:** Comment — the algorithm and tests are sound and well-documented; the issues are minor (unused import, doc overstatement, a residual-ignoring exit branch, and missing output-level regression evidence) and do not warrant blocking.

PR #134, opened 2026-08-14
Fix parallel transport frequency race

zandivx:
Add missing threadprivate and AI-generated tests if paralle execution works as intended.

### DIFF
--- a/src/transport.f90
+++ b/src/transport.f90
@@ -17,7 +17,10 @@ module neort_transport
 
   implicit none
 
+  ! These values remain live while bounce_fast invokes timestep_transport.
+  ! Each transport worker needs an independent callback state.
   real(dp) :: Omth, dOmthdv, dOmthdeta
+  !$omp threadprivate (Omth, dOmthdv, dOmthdeta)
 
 contains
 
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -27,6 +27,18 @@ add_test(NAME test_neort_lib
     WORKING_DIRECTORY ${CMAKE_SOURCE_DIR}/examples/base
 )
 
+add_test(NAME test_parallel_transport
+    COMMAND ${CMAKE_BINARY_DIR}/test_parallel_transport.x
+    WORKING_DIRECTORY ${CMAKE_SOURCE_DIR}/examples/base
+)
+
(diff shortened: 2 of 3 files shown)


krystophny (approved):
**Review verdict: Approve**

## Review of PR #134: Fix parallel transport frequency race

**Summary:**
This PR fixes a genuine OpenMP data race in `src/transport.f90` by marking the module-level working variables `Omth, dOmthdv, dOmthdeta` as `!$omp threadprivate`, and adds a new regression test (`test/test_parallel_transport.f90`) wired into CTest. The race is real: these three variables are written per-flux-surface in `compute_transport_integral` (via `Om_th`) and read asynchronously inside the `bounce_fast`/VODE callback `timestep_transport` and in `nonlinear_attenuation`. With the pre-existing `test_neort_lib.f90` already calling `neort_compute_at_s` concurrently over flux surfaces, a shared (non-threadprivate) global could be overwritten by another thread mid-integration, corrupting `taub`, `Hn`, and the transport/torque results. The change is minimal, correct, and consistent with the established codebase pattern where every other per-thread working variable (`s`, `q`, `mth`, `etatp`, `B0`, etc.) is already threadprivate. The remaining globals read in the callback are read-only config (`epsmn`, `m0`, `pertfile`, `nonlin`), so the fix is complete for that path. I could not build here (`examples/base` is not present in the working tree), so no local execution evidence; the test is however correctly wired and CI (`make ctest`, Fast+Debug) exercises it.

**Findings:**
1. [minor] test/test_parallel_transport.f90:35-51 — The regression test is inherently probabilistic: with the race present, results may still match if scheduling happens not to interleave `Om_th` writes with callback reads. `N_REPEATS=4` with `schedule(dynamic,1)` across 8 surfaces helps but does not guarantee a failure on the unfixed code, so a false pass is possible. Not blocking; consider documenting the probabilistic nature or raising repeats.
2. [minor] src/transport.f90:22-23 — The fix relies on `Omth` being set and consumed on the same OpenMP thread. The inline comment explains the callback dependency well; a short note that threadprivate now enforces this invariant would help future callers. Cosmetic.

**Verdict:** Approve — the fix is minimal, correct, and consistent with the codebase's established threadprivate pattern, and the new test is well-structured and wired into CI; the only caveats are the inherent probabilistic nature of race-detection tests, which do not block merging.

PR #135, opened 2026-08-17
Split analytic and file perturbation scaling

krystophny:
## Summary

- Keep `epsmn` for analytic perturbations (`pertfile=.false.`) and set its default to 1.
- Add `pertfile_scale` for perturbations read from `in_file_pert` (`pertfile=.true.`), defaulting to 1.
- Apply the separate scale in the existing magnetic-field and transport paths.
- Update the public configuration docs, examples, and the perturbation behavior probe.

## Validation

- `ctest --test-dir build --output-on-failure` — 14/14 passed.
- `test_perturbation_fourier_kernel` — passed.
- `git diff --check` — passed.

### DIFF
--- a/doc/library.md
+++ b/doc/library.md
@@ -134,7 +134,7 @@ Computes transport using `s`, `M_t`, `vth`, etc. from the config file or `config
 type :: config_t
     ! Same fields as the params namelist in <base>.in
     real(8) :: s, qs, ms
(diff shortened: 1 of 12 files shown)


krystophny (commented):
**Review verdict: Comment**

See review below.

zandivx:
@krystophny please make sure the actions pass

